PR #16, opened 2020-12-21
updates rc

roninsightrx:


mattfidler:
Travis is going away by the end of the year.  You may want to check out github actions

PR #17, opened 2021-09-28
RXR-387: bring in updates from PKPDsim2

karawoo:
This brings in all commits from our internal version of PKPDsim2, making the public version fully up to date.


Jenkinsfile
```diff
@@ -0,0 +1,52 @@
...
+        sh """
+        \$(aws ecr get-login --no-include-email --region us-west-2 &> /dev/null)
+        docker run -d --name ${BUILD_TAG} 579831337053.dkr.ecr.us-west-2.amazonaws.com/irx-r-base:5
+        """
+      }
+
+    }
+    stage('Build & test PKPDsim') {
+      steps {
+        echo 'Installing and checking irxtools'
```
jasmineirx, line 24: this label is incorrect, should read 
```suggestion
        echo 'Installing and checking PKPDsim'
```

DESCRIPTION
```diff
@@ -1,16 +1,18 @@
...
+Date: 2021-06-08
 Author: Ron Keizer
-Maintainer: Ron Keizer <ronkeizer@gmail.com>
+Maintainer: Ron Keizer <ron@insight-rx.com>
 Depends: R (>= 3.0.2)
-Imports: Rcpp (>= 0.12.9), BH, dplyr, data.table, lubridate, stringr,
-        MASS, randtoolbox, jsonlite, stats, httr
+Imports: Rcpp (>= 0.12.9), BH, data.table, lubridate, stringr, MASS,
+        randtoolbox, jsonlite, stats, parallel
+Suggests: httr, testit, testthat (>= 3.0.0)
```
jasmineirx, line 11: do we still need testit? (can be a separate pr)

jasmineirx (approved):
lgtm! Since the LOC changes have previously been reviewed, I considered mainly the new files related to infrastructure (ex, jenkinshile, slack notifications) and they look good to release.

Exciting!

roninsightrx (approved):
👍🏻 nice

DESCRIPTION
karawoo: I think you're right, we don't need it. Will remove.

karawoo:
I set up a Jenkins project for this repo so I'll double-check that it succeeds before merging. I also removed the `.travis.yml` file, I don't think we are using travis anymore (and really shouldn't be given its recent security flaws).

PR #18, opened 2021-11-09
RXR-597 fix handling of multiple observation types

roninsightrx:
it was working fine for no observation types, or <= 2 observation types, but not for >2. There was an issue compiling the C++ code, as well creating the model attributes.

### DIFF
--- a/R/compile_sim_cpp.R
+++ b/R/compile_sim_cpp.R
@@ -223,8 +223,16 @@ compile_sim_cpp <- function(
         if(length(obs$scale) < length(obs$variable)) {
           obs$scale <- rep(obs$scale[1], length(obs$variable))
         }
-        tmp <- paste0("      if (obs_type[i+1]==", 1:length(obs$variable), ") { obs.insert(obs.end(), ", obs$variable, "/", obs$scale, "); } else ")
-        tmp <- c(tmp, paste0("        { obs.insert(obs.end(), ", obs$variable[1], "/", obs$scale[1], "); }")) # make sure something is pushed on obs stack
+        tmp <- c()
+        for(i in 1:length(obs$variable)) {
+          str_else <- ifelse(length(obs$variable)-i <= 0, str_else <- "else", "else if")
+          str_if <- ifelse(i == 1, "if", "  ")
+          tmp <- c(
+            tmp, 
+            paste0("      ", str_if, " (obs_type[i+1]==", i, ") { obs.insert(obs.end(), ", obs$variable[i], "/", obs$scale[i], "); } ", str_else, " ")
+          )
+        }
+        tmp <- c(tmp, paste0("         { obs.insert(obs.end(), ", obs$variable[1], "/", obs$scale[1], "); }")) # make sure something is pushed on obs stack
       }
       cpp_code[idx12] <- paste0(tmp, collapse = "\n")
     }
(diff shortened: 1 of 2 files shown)


jasmineirx (changes requested):
the approach looks okay, but I think we should wrap this into a function and add a unit test. The logic is well contained and only takes a couple pieces of information.

tests/testthat/test_parse_obs_types.R
```diff
@@ -0,0 +1,46 @@
...
+})
+
+test_that("multiple observation type, scale is single", {
+  obs3a <- list(
+    variable = c("CONC", "CONCM", "PD"),
+    scale = 1
+  )
+  expect_warning(
+    res <- PKPDsim:::parse_obs_types(obs3a)
+  )
```
jasmineirx, line 29:
```suggestion
  expect_warning(
    res <- parse_obs_types(obs3a),
    "Provided `scale` vector not same length as `variable` vector."
  )
```
if i'm correct this is a warning we are raising? might want to check for it

R/parse_obs_types.R
```diff
@@ -0,0 +1,25 @@
...
+      warning("Provided `scale` vector not same length as `variable` vector.")
+    }
+    tmp <- c()
+    for(i in 1:length(obs$variable)) {
+      str_else <- ifelse(length(obs$variable)-i <= 0, str_else <- "else", "else if")
+      str_if <- ifelse(i == 1, "if", "  ")
+      tmp <- c(
+        tmp,
+        paste0("      ", str_if, " (obs_type[i+1]==", i, ") { obs.insert(obs.end(), ", obs$variable[i], "/(", obs$scale[i], ")); } ", str_else, " ")
+      )
```
jasmineirx, line 20:
```suggestion
        paste0(
          "      ", str_if, " (obs_type[i+1]==", i, ")", 
          " { obs.insert(obs.end(), ", obs$variable[i], "/(", obs$scale[i], ")); } ", 
          str_else, " "
        )
      )
```
adding line breaks like this might make it a bit easier to read - tried to mimic how line breaks might look if I were writing it in code.

R/parse_obs_types.R
```diff
@@ -0,0 +1,25 @@
...
+    }
+    tmp <- c(tmp, paste0("         { obs.insert(obs.end(), ", obs$variable[1], "/(", obs$scale[1], ")); }")) # make sure something is pushed on obs stack
```
jasmineirx, line 22:
```suggestion
    # make sure something is pushed on obs stack
    tmp <- c(
      tmp, 
      paste0("         { obs.insert(obs.end(), ", obs$variable[1], "/(", obs$scale[1], ")); }")
    ) 
```
this line is a bit long, suggest adding line breaks and moving the comment to its own line

R/parse_obs_types.R
```diff
@@ -0,0 +1,25 @@
+#' Parse observation types to simulation code
+#'
+#' @param obs specified observation object, e.g. `list(variable="CONC", scale="1")` or `list(cmt=1, scale="V/1000.0")`.
```
jasmineirx, line 3: is the second example valid here? This function will only get called if the following is true:

```
!is.null(obs$variable) && length(obs$variable) > 0
```
Yet the second example is missing a `variable` object. I suggest clarifying the expectations in the documentation.

```suggestion
#' @param obs specified observation object including at least a description of 
       which variable(s) are associated with a particular compartment, e.g. 
       `list(variable="CONC", scale="1")`.
```

jasmineirx (approved):
lgtm, a couple style comments

PR #19, opened 2021-11-17
RXR-602 fix issue lag time for models with multiple compartments

roninsightrx:
A bug was present when lagtime was specified as a parameter in a model (e.g. TLAG), and the lag time definition specified as a vector for the compartments, e.g. `["TLAG", 0, 0]`. This caused the lag-time not to be applied. This PR fixes that behavior and updates unit tests to include that scenario.


karawoo (approved):
lgtm

PR #20, opened 2021-11-17
RXR-602 more fixes to lagtime implementation

roninsightrx:
We have to supply cmt_mapping that is tied to the model as well, otherwise lag times are not supplied to the correct compartments


karawoo (approved):
lgtm, tested against the models repo and other than the one case addressed by https://github.com/InsightRX/models/pull/356 things seem to be working there now 👍

PR #21, opened 2021-11-19
RXR-616: fix bug multiple infusion stop events

roninsightrx:
When a simulation was performed with multiple observation types *and* on of the specified observation times coincided with an infusion stop event, the infusion stop event would be repeated due to a merge action. This resulted in negative infusion rates.

### DIFF
--- a/R/sim.R
+++ b/R/sim.R
@@ -387,7 +387,7 @@ sim <- function (ode = NULL,
     p_i <- p
     if(!is.null(covariates_table)) {
       covariates_tmp <- covariates_table[[i]]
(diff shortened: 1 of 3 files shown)


R/sim.R
```diff
@@ -387,7 +387,7 @@ sim <- function (ode = NULL,
     p_i <- p
     if(!is.null(covariates_table)) {
       covariates_tmp <- covariates_table[[i]]
-      design_i <- create_event_table(regimen, t_max, t_obs, t_tte, t_init = t_init, p_i, covariates_table[[i]])
```
roninsightrx, line 390: unrelated, but noticed that this instance of calling create_event_table did not have all arguments, which could in specific cases lead to errors.

karawoo (approved):
lgtm

PR #22, opened 2021-12-07
RXR-639 reduce package size (strip libs)

roninsightrx:
Requires trick described here:

http://dirk.eddelbuettel.com/blog/2017/08/14/#009_compact_shared_libraries
https://stackoverflow.com/questions/46280628/object-files-in-r-package-too-large-rcpp

With the Makevars file, the NOTE about pkg size is not thrown anymore.


karawoo (approved):
Nice, thanks!

PR #23, opened 2021-12-07
RXR-634: PKPDsim cleanup

karawoo:
Addresses some NOTEs and other minor issues that we need to clean up before submitting to CRAN, including

- Naming everyone on the DS team as contributors
- Updating the license to the correct year and naming InsightRX as copyright holder
- Removing subdirectories from `R/`
- Removing broken or redirected links (checking with `_R_CHECK_CRAN_INCOMING_REMOTE_` set to true, which CRAN does, checks that links aren't broken and don't redirect)
- Removing `sim_ae()`


R/sim_core_analytic.R
```diff
@@ -2,16 +2,17 @@
 #' Mostly useful for estimations / optimal design. Has no checks (for speed)!
 #' This is separate function for analytic equations only. See `sim_core()` for ODE function.`
 #'
+#' @inheritParams sim
 #' @param sim_object list with design and simulation parameters
 #' @param model model object
 #' @param duplicate_t_obs allow duplicate t_obs in output? E.g. for optimal design calculations when t_obs = c(0,1,2,2,3). Default is FALSE.
 #' @param t_init time of initization of the ODE system. Usually 0.
 #' @export
-sim_core_analytic <- function(
```
karawoo, line 10: Added `ode` as an argument here, otherwise r cmd check complains that it's not found. `sim()` takes `ode` as an arg so I guessed it should be the same here, but not certain.

src/RcppExports.cpp
```diff
@@ -5,6 +5,11 @@
 
 using namespace Rcpp;
 
+#ifdef RCPP_USE_GLOBAL_ROSTREAM
+Rcpp::Rostream<true>&  Rcpp::Rcout = Rcpp::Rcpp_cout_get();
+Rcpp::Rostream<false>& Rcpp::Rcerr = Rcpp::Rcpp_cerr_get();
+#endif
+
```
karawoo, line 12: When I check the package locally it automatically adds these lines. I'm not sure if they are strictly necessary to commit or not, though.

R/sim_core_analytic.R
```diff
@@ -2,16 +2,17 @@
...
-  sim_object = NULL,
-  model,
-  duplicate_t_obs = FALSE,
-  t_init = 0) {
+sim_core_analytic <- function(ode,
```
karawoo, line 11: Added `ode` as an argument here, otherwise r cmd check complains that it's not found. `sim()` takes `ode` as an arg so I guessed it should be the same here, but not certain.

DESCRIPTION
```diff
@@ -1,18 +1,23 @@
...
-Maintainer: Ron Keizer <ron@insight-rx.com>
+Date: 2021-12-06
+Authors@R: c(
+        person("Ron", "Keizer", email = "ron@insight-rx.com", role = c("aut", "cre")),
+        person("Jasmine", "Hughes", email = "jasmine@insight-rx.com", role = "aut"),
+        person("Dominic", "Tong", email = "dominic@insight-rx.com", role = "aut"),
+        person("Kara", "Woo", email = "kara@insight-rx.com", role = c("aut")),
+        person("InsightRX", role = c("cph", "fnd")))
 Depends: R (>= 3.0.2)
-Imports: Rcpp (>= 0.12.9), BH, data.table, lubridate, stringr, MASS,
```
karawoo, line 9: Removing BH from Imports and moving it to LinkingTo generates a NOTE on CI, but having it the other way around generates a NOTE locally. I believe this is due to CI using an older version of Rcpp and that it's correct to move it to LinkingTo.

src/RcppExports.cpp
roninsightrx: this happens to me as well, have been manually deleting thus far to look at later. But I think it's fine to commit.

R/sim_core_analytic.R
roninsightrx: hmm, that is a bit surprising. I'm ok with committing this for now, will look at this when I get to RXR-633 later this week / next.

roninsightrx (approved):
lgtm

DESCRIPTION
karawoo: Argh I was wrong, this is actually an error. We might need to update the irx-r-base docker image to use newer package versions, at least for Rcpp.

karawoo: Updating the Rcpp version doesn't seem to fix it, so I am reverting this change for now and will look into it separately.

PR #24, opened 2021-12-08
RXR-635: Skip slow tests on CRAN

karawoo:
Some of our tests are relatively slow, often due to calls to `new_ode_model()`. This skips some tests on CRAN and rearranges some of the test code to put it within `test_that()` calls (so it can be easily skipped). In `test_advan_with_auc.R` there's a lot of code outside of `test_that()` calls that gets reused in multiple tests, so rather than duplicate it I put it in an `if` clause that checks for the `NOT_CRAN` environment variable. This is the same approach taken by `skip_on_cran()`, so I think it should be reasonable.

On my machine, `devtools::check(env_vars = c("NOT_CRAN" = "false"))` reports that the test suite takes 54 seconds. It's possible results will be different on CRAN machines depending on their load, but there's no real way to be certain until we try, so I think this is good enough for now.

### DIFF
--- a/tests/testthat/test_calc_ss_analytic.R
+++ b/tests/testthat/test_calc_ss_analytic.R
@@ -8,6 +8,9 @@ reg_inf <- new_regimen(amt = dose, interval = interval, n = n_ss, type = "infusi
 t_obs <- max(reg_oral$dose_times) + interval
 # Uses models defined in setup.R
 
+if (identical(Sys.getenv("NOT_CRAN"), "true")) {
+  pk_3cmt_oral <- new_ode_model("pk_3cmt_oral")
+}
 
 #delta <- function(x, ref) { abs(x-ref)/ref }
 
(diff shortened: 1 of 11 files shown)


jasmineirx:
What's our best process for adding tests to this package as we extend it? 54s doesn't give us a lot of wiggle room 😅 Does it make sense to (in a separate PR) add a CI check that looks at test suite duration with the CRAN flag on? (Is that something people do?)

tests/testthat/test_calc_ss_analytic.R
```diff
@@ -94,6 +94,8 @@ test_that("2-cmt infusion", {
 })
 
 test_that("3-cmt oral", {
+  skip_on_cran()
+  pk_3cmt_oral <- new_ode_model("pk_3cmt_oral")
```
jasmineirx, line 98: we are building this model (and the iv version) twice now, which makes our testing a little slower when CRAN is set to false. Can we bring it out of the test_that chunks and wrap it in the `if (identical(Sys.getenv("NOT_CRAN"), "true")) { ...}` check that you used elsewhere?

jasmineirx (approved):
lgtm, one suggestion for speed for non-cran tests

karawoo:
Agree it's not a lot of wiggle room, but testing the time it takes seems a bit overkill, and likely to be different on different systems anyway.

I got the 1 minute guideline from the [R Packages](https://r-pkgs.org/tests.html#test-cran) book, and re-reading it I'm not actually sure that the whole test suite needs to be under 1 minute -- maybe it's referring to individual tests? The [CRAN policies](https://cran.r-project.org/web/packages/policies.html) actually only say that "Checking the package should take as little CPU time as possible". I'm inclined to wait until we hear anything definitive from them before trying to optimize further, what do you think?

jasmineirx:
The proposed approach works for me -- we are still testing a lot of the core functionality, just not some of the mroe niche features.

PR #25, opened 2021-12-16
RXR-636: CRAN submission prep

karawoo:
Updates the repo based on some of the things listed in [extrachecks](https://github.com/DavisVaughan/extrachecks). The main thing was adding `@return` documentation for every exported function. I also tried to update the package title and description to be a little more informative, and changed the `sim()` example so that it does not use ggplot2 (since we don't list ggplot2 as a suggested dependency in the DESCRIPTION).

I did not add return value documentation for `analytical_eqn_wrapper()` or `sim_core_analytic()` until we decide if we're keeping those per RXR-633.

### DIFF
--- a/R/na_locf.R
+++ b/R/na_locf.R
@@ -7,7 +7,7 @@
 #'   than forward. Default is FALSE.
 #'
 #' @export
-#' @return Original object iwth NAs filled in
+#' @return Original object with NAs filled in
 na_locf <- function(object, fromLast = FALSE) {
   if (fromLast) object <- rev(object)
   for (i in seq_along(object)) {
--- a/R/na_locf.R
+++ b/R/na_locf.R
@@ -3,9 +3,11 @@
 #' Inspired by zoo::na.locf0
 #'
 #' @param object an object
-#' @param fromLast logical. Causes observations to be carried backward rather than forward. Default is FALSE. With a value of TRUE this corresponds to NOCB (next observation carried backward). It is not supported if x or xout is specified.
+#' @param fromLast logical. Causes observations to be carried backward rather
+#'   than forward. Default is FALSE.
 #'
 #' @export
+#' @return Original object iwth NAs filled in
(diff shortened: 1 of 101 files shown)


R/na_locf.R
```diff
@@ -3,9 +3,11 @@
 #' Inspired by zoo::na.locf0
 #'
 #' @param object an object
-#' @param fromLast logical. Causes observations to be carried backward rather than forward. Default is FALSE. With a value of TRUE this corresponds to NOCB (next observation carried backward). It is not supported if x or xout is specified.
+#' @param fromLast logical. Causes observations to be carried backward rather
+#'   than forward. Default is FALSE.
 #'
 #' @export
+#' @return Original object iwth NAs filled in
```
roninsightrx, line 10:
```suggestion
#' @return Original object with NAs filled in
```

R/new_covariate.R
```diff
@@ -10,6 +10,8 @@
 #' @param comments `NULL`, or vector of length equal to `value` specifying comments to each observation
 #' @param verbose verbosity
 #' @export
+#' @return Object of class `"covariate"
```
roninsightrx, line 13:
```suggestion
#' @return Object of class `"covariate"`
```

R/sim_core.R
```diff
@@ -6,6 +6,7 @@
 #' @param duplicate_t_obs allow duplicate t_obs in output? E.g. for optimal design calculations when t_obs = c(0,1,2,2,3). Default is FALSE.
 #' @param t_init time of initization of the ODE system. Usually 0.
 #' @export
+#' @return Data frame of observations
```
roninsightrx, line 9:
```suggestion
#' @return Data frame with simulation results
```

roninsightrx (changes requested):
looks good, just a few suggestions for improvement of description and typo-fixes

roninsightrx (approved):
👍🏻

PR #26, opened 2021-12-20
RXR-633 rm unused functionality

roninsightrx:
This removes an earlier implementation of analytic equations, since we now have a newer advan() function that handles that. While sim_core_analytic() is in principle a somewhat more generalized version of advan(), in practice I don’t think there will be a use case for it. I checked that neither sim_core_analytic() nor analytical_eqn_wrapper() is used in any of our other R packages.


karawoo (changes requested):
I believe we need to update this line:
https://github.com/InsightRX/PKPDsim/blob/3af74cf588c4e256a9e1cac0bb66b483b97fdf0d/R/sim.R#L469

roninsightrx:
> I believe we need to update this line:
> 
> https://github.com/InsightRX/PKPDsim/blob/3af74cf588c4e256a9e1cac0bb66b483b97fdf0d/R/sim.R#L469

ah darn, then we can't analytical_eqn_wrapper(). Will rollback partially.

karawoo:
If we're only using it internally maybe let's not export it?

karawoo (approved):

PR #27, opened 2021-12-21
RXR-637: fix note about lubridate and randtoolbox

karawoo:
Removes lubridate dependency in favor of a base R implementation of the one function we were using. Also updates our use of randtoolbox functions so that they are detected by R CMD check. Both updates are to avoid this note that has been showing up on rhub (though not locally for some reason).

Screen Shot 2021-12-20 at 2 48 08 PM


roninsightrx (approved):
👍🏻

PR #28, opened 2022-01-07
CRAN submission

roninsightrx:


R/new_ode_model.R
```diff
@@ -404,6 +404,7 @@ new_ode_model <- function (model = NULL,
 
       ## Compile / build / install
       curr <- getwd()
+      on.exit(setwd(curr))
       setwd(new_folder)
```
karawoo, line 408: Do we actually need `setwd()` here? It seems like we can put files in the correct place based on the `new_folder` path without changing the working directory, but maybe I'm missing where it's needed.

roninsightrx: I think it's about the R CMD INSTALL command that is (optionally) run, about 15 lines below the setwd() command. If I remember correctly I wasn't able to install the new package without actually being in that folder...

karawoo: I just tried it and if we change `Rcpp::compileAttributes(".", )` to `Rcpp::compileAttributes(new_folder, )` a few lines down it seems to work without needing `setwd()`. I think @jasmineirx's changes in https://github.com/InsightRX/PKPDsim2/pull/68 fixed the issue with R CMD INSTALL that was present before.

roninsightrx: OK, thx!

karawoo (approved):
LGTM

PR #29, opened 2022-01-20
RXR-792: updates to allow double absorption pazopanib

roninsightrx:
Some absorption models are implemented using double dosing compartments with different absorption rates, each possibly having a lag time as well. Example is the pazopanib model by Yu et al, but I have seen it before in other models as well. Some adjustments to the regimen have to be made, essentially just a duplication of the regimen and adjustment of the compartments. However, we should implement a feature so this can be defined in model metadata, and ensure no other action is required in module metadata (since this is really a "model" feature and not a "module" feature).

For the pazopanib model specifically, a separate other fix in PKPDsim is required too: the model defines the bioavailability of the absorption compartments using the previous dose:
TVB  = (1-(prv_dose-200)*EMAX/(ED50+prv_dose-200))
However, this fails for the first dose, because the `prv_dose` variable is only initiated after the first row of the event record during simulation. So the correct value for biovailability does not get applied during the first dose. This is a simple fix in sim.cpp.


inst/cpp/sim.cpp
```diff
@@ -76,7 +76,8 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
   // insert_parameter_definitions
 
   // Initialize parameters, compartments, etc:
-  pk_code(0, times, doses, 0, dose_cmt, dose_type, iov_bin);
+
+  pk_code(0, times, doses, doses[0], dose_cmt, dose_type, iov_bin);
```
roninsightrx, line 80: this fix, and the one a few lines below, are required to make sure `prv_dose` has the correct value at the first row of the event table for the simulation.

R/compile_sim_cpp.R
```diff
@@ -122,8 +122,12 @@ compile_sim_cpp <- function(
     pars <- paste0(pars, "double ", p[i], ";\n")
   }
   pars <- paste0(pars, "double prv_dose, t_prv_dose = 0;\n")
+  ## Initialize rate and bioav variables and insert temporary dummy values.
   pars <- paste0(pars, paste0("double rate[] = { ", paste(rep(0, size), collapse=", "), " };\n"))
-  pars <- paste0(pars, "double bioav = 1;\n")
+  if(length(dose$bioav) != size) {
```
roninsightrx, line 127: bioavailability has to be a vector now

R/merge_regimen.R
```diff
@@ -9,6 +9,7 @@
 #' @return Merged regimens
 merge_regimen <- function(regimens) {
   cols <- c("interval", "n", "type", "t_inf", "dose_times", "dose_amts", "first_dose_time")
+  if("cmt" %in% names(regimens[[1]])) { cols <- c(cols, "cmt") }
```
roninsightrx, line 12: only merge `cmt` values if present. This is a bit ugly to have to include. But the way we implemented is that the regimen is in principle agnostic about the model, so we are allowed to define regimens without explicitly stating the `cmt` it will go in, and then rely on defaults or the `cmt_mapping` variable. However, in the case of double absorption, we have to be explicit about the compartments.

tests/testthat/test_new_ode_model.R
```diff
@@ -1,3 +1,37 @@
+test_that("Double absorption models work appropriately", {
```
jasmineirx, line 1:
```suggestion
test_that("Double absorption models work appropriately", {
  skip_on_cran()
```

tests/testthat/test_new_ode_model.R
```diff
@@ -1,3 +1,37 @@
...
+    only_obs = F,
+    output_include = list(variables=TRUE),
+    t_obs = seq(0, 24, .1)
+  )
+  expect_equal(attr(mod, "dose")$duplicate, 2)
+  expect_equal(round(res[res$comp == 1 & res$t == 0.1, ]$y, 2), 27.15) # abs comp 1
+  expect_equal(res[res$comp == 2 & res$t == 1.2, ]$y, 0) # abs comp 2: should still be 0 due to lagtime
+  expect_equal(round(res[res$comp == 2 & res$t == 1.6, ]$y, 1), 54.3) # abs comp 2: should be twice the amount of t=0.1&comp==1
+})
+
```
jasmineirx, line 34: Should we also have an example where we need to rely on cmt_mapping to associate the dose type with the right cmt (i.e., a mix of oral and infusion dosing) and use a 2-cmt absorption model?

jasmineirx: Maybe we can also have a test where the bioavailability is supplied as an expression? e.g.: `"(dose_cmt[i] == 1) ? Fi : 1.0"`

jasmineirx (commented):
in principle the approach looks good to me; it particularly makes sense to initialize prv_dose and t_prv_dose.

a couple questions about edge cases/interactions that i think we should add tests for.

tests/testthat/test_new_ode_model.R
roninsightrx: As discussed on Slack, we should not use that syntax anymore, and update the models that use it.

jasmineirx (approved):
lgtm, but can we add a test for `merge_regimen` since we changed the behavior there too?

PR #30, opened 2022-01-24
RXR-799: parse bioav as array

jasmineirx:
In https://github.com/InsightRX/PKPDsim/pull/29 we made changes to allow for multiple absorption compartments, which necessitated allowing bioavailability to be specified as a vector. This PR extends `new_ode_model` to allow bioavailability fields in model JSON files (or other methods of specifying a model) to be provided as a vector.

Only one line really needed to be changed (see commit https://github.com/InsightRX/PKPDsim/pull/30/commits/43db18f9a1bde3fa1a67492bd906f67a0cef35f4); the rest of the lines of code changes were to enable better unit testing and readability.


tests/testthat/test_object_to_R_code.R
```diff
@@ -0,0 +1,41 @@
...
+  expect_true(grepl("\\)$", x_empty))
+})
+
+test_that("Express bioavailability as parseable R", {
+  bioav_null <- bioavailability_to_R_code(NULL)
+  expect_equal(eval_string(bioav_null), "1")
+  expect_true(grepl("^c\\(", bioav_null))
+  expect_true(grepl("\\)$", bioav_null))
+
+  vec_char <- c("F1i", "1")
```
roninsightrx, line 29: What about a test for the ternary-style definition, if the plan is to still support that?

roninsightrx (approved):
I thought when you said on Friday you would implement fixes that you would just update the models that used ternary specifications to use the new syntax. But this is fine too, so we keep the flexibility. Or wasn't that your plan?
LGTM, just one q about test.

tests/testthat/test_object_to_R_code.R
jasmineirx: This change is required to allow for the new specification. It wasn't working without this. The next step is to update all the model files. I locally checked that this would work with one, I can push that up if you'd like to check?

jasmineirx: The intent is to no longer support ternary-style definitions.

R/object_to_R_code.R
```diff
@@ -0,0 +1,36 @@
...
+#' (and special characters escaped). A general purpose helper function
+#' for writing new model code.
+#'
+#' @param vec a vector
+#' @param return_null logical indicating if `vec` is NULL whether to "NULL"
+#'   (alternative is to return an empty list).
+#' @return character string of length 1
+
+vector_to_R_code <- function(vec, return_null = TRUE) {
+  if (return_null && is.null(vec)) return("NULL")
```
karawoo, line 13: Does the model code actually behave differently if we use `c()` vs. `NULL`, since both evaluate to `NULL`?

R/object_to_R_code.R
```diff
@@ -0,0 +1,36 @@
...
+#'
+#' @param bioav bioavailability specification, either NULL (assume a value of 1
+#'   in all compartments), a single value (assume it applies to all
+#'   compartments), or a vector of values.
+#' @return character string of length 1
+
+bioavailability_to_R_code <- function(bioav) {
+  if (is.null(bioav)) bioav <- "1"
+
+  if(class(bioav) == "character" || length(bioav) > 1) {
```
karawoo, line 30: I know it was like this in the original code but I think `inherits()` is safer since it will always return a single value.

```suggestion
if(inherits(bioav, "character") || length(bioav) > 1) {
```

R/object_to_R_code.R
```diff
@@ -0,0 +1,36 @@
...
+    paste0("c(", paste(add_quotes(bioav), collapse = ", "), ")")
+  } else {
+    bioav
+  }
```
karawoo, line 34: Why is it that we convert multiple numeric values to string, but if it's a single numeric value it stays numeric?

```r
bioavailability_to_R_code(1)
#> [1] 1
bioavailability_to_R_code(c(1, 1))
#> [1] "c(\"1\", \"1\")"
```

karawoo (commented):
Just a couple of questions

R/object_to_R_code.R
jasmineirx: I suppose we could break it up so that:
- if its a single character, escape quotes
- if its a vector of characters, wrap in `c()` and escape quotes
- if its a vector of numbers, wrap in `c()` and don't escape quotes

it's probably a little more clear what happens, but by the time you get to functioning C code I don't think it matters

karawoo: If string vs. numeric doesn't matter, maybe we should just make everything a string?

R/object_to_R_code.R
jasmineirx: hmmm, maybe not 🤔 I can't get it to fail in a few examples of models I've tried. Do we like `NULL` or `c()` better?

karawoo: If we do `c()` we can remove this line and just have the `paste0()` call, OTOH `NULL` seems a little more explicit/readable. I don't feel too strongly one way or the other.

tests/testthat/test_object_to_R_code.R
roninsightrx: OK thx. Yes, can you share one example that isn't working? I tried it for pazopanib and that worked fine.

roninsightrx: Ah I see, it now only works for vectors, not for character anymore.

jasmineirx: @roninsightrx here, you can pull this branch: https://github.com/InsightRX/models/pull/380. this PR makes this specification work.

karawoo (approved):
LGTM, I think you need to re-run `devtools::document()` though.

R/object_to_R_code.R
```diff
@@ -0,0 +1,35 @@
...
+
+bioavailability_to_R_code <- function(bioav) {
+  if (is.null(bioav)) bioav <- "1"
+  bioav <- as.character(bioav)
+
+  if(length(bioav) > 1) {
+    vector_to_R_code(bioav)
+  } else {
+    add_quotes(bioav)
+  }
```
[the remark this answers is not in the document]
jasmineirx: i wasn't sure if it was confusing since it isnt necessarily a vector, but it's true that `c(1) == 1`

PR #31, opened 2022-01-25
RXR-792: fix placement of var init

roninsightrx:
In the previous PR (https://github.com/InsightRX/PKPDsim/pull/29/files), initialization of `prv_dose` and `t_prv_dos` was added, but by mistake this initialization was placed inside the main loop over the event table, and not before. This PR fixes that.

This would only have affected models that use `t_prv_dose` or `prv_dose` in their model code.


jasmineirx (approved):
lgtm, makes sense

PR #33, opened 2022-02-01
slight re-order of C++ code in main loop to make sure IOV on bioavailability is correctly applied

roninsightrx:
Because of the order of code in the sim.cpp, IOV defined in the PK block wasn’t called at the right times to actually affect biovailability properly. Requires a small change of order of code execution in the main loop in sim.cpp. 

Check that it doesn’t affect other models: in progress
Add a unit test: haven't been able yet to create a unit test that catches the error.

### DIFF
--- a/inst/cpp/sim.cpp
+++ b/inst/cpp/sim.cpp
@@ -96,21 +96,22 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
 
     set_covariates(i);
 
+    if(evid[i] == 1) {
+      t_prv_dose = times[i];
+      prv_dose = doses[i];
+    }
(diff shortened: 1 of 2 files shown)


karawoo:
@roninsightrx is this ready for review now or did you want me to wait until you've confirmed it doesn't affect other models?

roninsightrx:
> @roninsightrx is this ready for review now or did you want me to wait until you've confirmed it doesn't affect other models?

Sorry planned to do this earlier. Will report back within the hours if/how it affect modvals.

roninsightrx:
All models still pass (except pk_gentamicin_germovsek, but I believe we already singled that one out as needing some work anyhow).

roninsightrx:
> lgtm, I also ran insightrxr f-tests locally with this change and everything passed

Thx. I managed to implement a unit test for the issue.

PR #35, opened 2022-02-03
RXR-837 fix order output for models with lagtime

roninsightrx:
For some models with lagtimes for oral drugs, the amounts in compartment 1 will show a spike, due to the erroneous re-ordering of the output after simulation. The cause is a merge() command that altered the order of the output from the simulater. No “wrong” concentrations were simulated for any model, it’s just the output that was in some cases not arranged correctly.

### DIFF
--- a/R/sim.R
+++ b/R/sim.R
@@ -573,7 +573,7 @@ sim <- function (ode = NULL,
   comb$t_obs_type <- paste(comb$t, comb$obs_type, sep = "_")
   comb$rownr <- 1:nrow(comb) # keep row-ordering during merge
   comb <- merge(grid, comb, all=FALSE)[, c("id", "t_obs_type", "t", "comp", "y", "obs_type", "rownr", all_names)]
-  comb <- comb[order(comb$id, comb$comp, comb$t, comb$obs_type, comb$rownr, decreasing=FALSE), -2]
+  comb <- comb[order(comb$id, comb$comp, comb$t, comb$obs_type, comb$rownr, decreasing=FALSE), -c(2, 7)]
   if(!is.null(regimen_orig$ss_regimen)) {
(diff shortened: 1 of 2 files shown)


R/sim.R
```diff
@@ -571,8 +571,9 @@ sim <- function (ode = NULL,
   grid <- expand.grid(paste(t_obs, obs_type, sep="_"), unique(comb$id), unique(comb$comp))
   colnames(grid) <- c("t_obs_type", "id", "comp")
   comb$t_obs_type <- paste(comb$t, comb$obs_type, sep = "_")
-  comb <- merge(grid, comb, all=FALSE)[, c("id", "t_obs_type", "t", "comp", "y", "obs_type", all_names)]
-  comb <- comb[order(comb$id, comb$comp, comb$t, comb$obs_type, decreasing=FALSE), -2]
+  comb$rownr <- 1:nrow(comb) # keep row-ordering during merge
+  comb <- merge(grid, comb, all=FALSE)[, c("id", "t_obs_type", "t", "comp", "y", "obs_type", "rownr", all_names)]
+  comb <- comb[order(comb$id, comb$comp, comb$t, comb$obs_type, comb$rownr, decreasing=FALSE), -c(2, 7)]
```
jasmineirx, line 576: maybe it's a little clearer and safer to use column names just in case we change one of the two dataframes and the ordering changes? (also allows you to read exactly what is being removed)

```suggestion
  rm_cols <- which(colnames(comb) %in% c("t_obs_type", "rownr"))
  comb <- comb[order(comb$id, comb$comp, comb$t, comb$obs_type, comb$rownr, decreasing=FALSE), -rm_cols]
```

jasmineirx (approved):
lgtm!

karawoo (approved):

PR #36, opened 2022-02-09
RXR-794: Support scale as a vector

karawoo:
We were getting a warning in the tests that `scale` was not the same length as the `variable` vector, however if we extended `scale` to match the length we got compilation errors from the C++

```
filec0b122877146.cpp:146:5: error: use of undeclared identifier 'scale'
    scale = 1;
```

This has been fixed by using `scale1`, `scale2`, etc. if we have multiple values. This change is enough to make the updated test pass, and might be enough for this ticket.

However, I think there may be a similar issue in the `else` block below my changes:

https://github.com/InsightRX/PKPDsim/blob/8b0126658da30f7e0d67490361bf25e307631d2c/R/compile_sim_cpp.R#L214-L223

We seem to assume that if `length(obs$cmt) > 1`, then `obs$cmt` and `obs$scale` should always have the same length. I'm not sure if that assumption is valid, but when I tried to adjust this code it became a bit of a rabbit-hole. Possibly this is out of scope for this ticket, but @roninsightrx let me know if you think this is something we should address too.


roninsightrx:
> We were getting a warning in the tests that `scale` was not the same length as the `variable` vector, however if we extended `scale` to match the length we got compilation errors from the C++
> 
> ```
> filec0b122877146.cpp:146:5: error: use of undeclared identifier 'scale'
>     scale = 1;
> ```
> 
> This has been fixed by using `scale1`, `scale2`, etc. if we have multiple values. This change is enough to make the updated test pass, and might be enough for this ticket.
> 
> However, I think there may be a similar issue in the `else` block below my changes:
> 
> https://github.com/InsightRX/PKPDsim/blob/8b0126658da30f7e0d67490361bf25e307631d2c/R/compile_sim_cpp.R#L214-L223
> 
> We seem to assume that if `length(obs$cmt) > 1`, then `obs$cmt` and `obs$scale` should always have the same length. I'm not sure if that assumption is valid, but when I tried to adjust this code it became a bit of a rabbit-hole. Possibly this is out of scope for this ticket, but @roninsightrx let me know if you think this is something we should address too.

Yes, think we should address too. Wouldn't it be sufficient to say e.g. 

```
if(length(obs$scale) != length(obs$cmt)) {
  obs$scale <- rep(obs$scale[1], length(obs$cmt))
  warning("...")
}
```
Or would that have unforeseen consequences elsewhere?

karawoo:
> ```
> if(length(obs$scale) != length(obs$cmt)) {
>   obs$scale <- rep(obs$scale[1], length(obs$cmt))
>   warning("...")
> }
> ```

Do you mean we would do this only in cases where `length(obs$cmt) > 1`, or all the time? I don't think we can do it all the time unless we require `cmt` to be explicitly included in the `obs` object. Right now you don't have to include `cmt` and it will default to 1. Then if you have multiple scales the lengths won't match and we'd be only taking the first one, thus losing support for `scale` as a vector.

roninsightrx:
I think it's actually OK to have hard stops if scale is not provided correctly. Pushing up a change, lmk what you think.

Also, we will need a change to some models in the `models` repo, because some use `scale` in their code and not `scale1`. Will make a ticket.

R/compile_sim_cpp.R
```diff
@@ -209,7 +209,18 @@ compile_sim_cpp <- function(
...
+      if(!is.null(obs$cmt)) { ## `cmt` is specified
+        if(length(obs$scale) != length(obs$cmt)) {
+          stop("Provided `scale` vector is not the same length as observation compartments (`cmt`) vector. Please update `obs` argument.")
+        }
+      } else { ## `variable` is used
+        if(length(obs$scale) != length(obs$variable)) {
+          stop("Provided `scale` vector is not the same length as observation `variable` vector. Please update `obs` argument.")
+        }
+      }
       for(k in 1:length(obs$cmt)) {
```
karawoo, line 224: If I read this right, we'd require either `cmt` or `variable`. But if `variable` is provided and `cmt` is null, then won't the for loop below (`for k in 1:length(obs$cmt)`) fail?

roninsightrx: Right. I think that whole for-loop could be moved into the if-statement on line 215. I.e. I don't think it's necessary to include this loop when the observation is specified using a variable instead of cmt. The only important grep label for `variable` is idx2, and that is handled on line 233-235. But not entirely sure, would need to test.

Another option is to move that for loop out of this block and one level up. I.e. add it as `else` to the if-statement on line 233-235?
```
if(!is.null(obs$variable) && length(obs$variable) > 0) {
  cpp_code[idx12] <- parse_obs_types(obs)
} else {
   for(k in 1:length(obs$cmt)) {
        cpp_code[idx5] <- paste0(cpp_code[idx5], "\n    scale", k," = ", obs$scale[k], ";")
        cpp_code[idx7] <- paste0(cpp_code[idx7], "\n      scale", k," = ", obs$scale[k], ";")
        cpp_code[idx8] <- cov_scale
        cpp_code[idx12] <- paste0(cpp_code[idx12], "\n      obs",k,".insert(obs",k,".end(), tmp.y[k][", obs$cmt[k]-1,"] / scale", k,");")
        cpp_code[idx13] <- paste0(cpp_code[idx13], '\n  comb["obs', k,'"] = obs', k,';');
        cpp_code[idx14] <- paste0(cpp_code[idx14], "\n  std::vector<double> obs",k,";")
  }
}
```

karawoo: Coming back to this, I think in most cases this won't behave as expected, because `new_ode_model()` alters the `obs` object before passing it to `compile_sim_cpp()`. See the below example where we have a scale/variable length mismatch and no `cmt`. We should expect an error, but we don't get one:

``` r
## define parameters
vars <- c("CONC", "METAB", "METAB2")
pk1 <- new_ode_model(
  code = "dAdt[1] = -(CL/V)*A[1]; CONC = 1000*A[1]/V; METAB = CONC/2; METAB2 = CONC * t;",
  obs = list(variable = vars, scale = c(1, 1)), # mismatch vars/scale length; no cmt
  declare_variables = vars,
  cpp_show_code = F
)
#> Warning in parse_obs_types(obs): Provided `scale` vector not same length as
#> `variable` vector.
```

That's because `new_ode_model()` adds a `cmt` element to the `obs` object, so we never make it to these error conditions.

https://github.com/InsightRX/PKPDsim/blob/08677e82f59e37882a2bd76a04d39b4d28fe849b/R/new_ode_model.R#L193

I think it might be a good idea to come up with a more comprehensive set of test cases for this behavior before making more changes to the code. Maybe a brief meeting would help?

karawoo: Coming back to this, I think in most cases this won't behave as expected, because `new_ode_model()` alters the `obs` object before passing it to `compile_sim_cpp()`. See the below example where we have a scale/variable length mismatch and no `cmt`. We should expect an error, but we don't get one:

``` r
## define parameters
vars <- c("CONC", "METAB", "METAB2")
pk1 <- new_ode_model(
  code = "dAdt[1] = -(CL/V)*A[1]; CONC = 1000*A[1]/V; METAB = CONC/2; METAB2 = CONC * t;",
  obs = list(variable = vars, scale = c(1, 1)), # mismatch vars/scale length; no cmt
  declare_variables = vars,
  cpp_show_code = F
)
#> Warning in parse_obs_types(obs): Provided `scale` vector not same length as
#> `variable` vector.
```

That's because `new_ode_model()` adds a `cmt` element to the `obs` object, so we never make it to these error conditions.

https://github.com/InsightRX/PKPDsim/blob/08677e82f59e37882a2bd76a04d39b4d28fe849b/R/new_ode_model.R#L193

I think it might be a good idea to come up with a more comprehensive set of test cases for this behavior before making more changes to the code. Maybe a brief meeting would help?

karawoo:
@roninsightrx I think this is ready for another look. As we discussed in the DS meeting, we will no longer support `variable` and `scale` at the same time. I've updated the code accordingly and pulled out some of the logic for validating the contents of `obs` into its own function.

roninsightrx (approved):
LGTM, I think this is the cleanest solution. 
I've created a ticket  / PR to update some of our models to adhere to the new restrictions. With those updates, all models install and models pass unit and functional tests. Would be best to merge that PR first, then this one.

PR #37, opened 2022-04-01
CRAN submit of 1.1.1

roninsightrx:
- Added some info for CRAN reviewer
- Added NEWS
- updated suggestion from checks to update the way classes are checked (using `inherit()`) 
- removed the stripping of the compiled files, since that seems to be a violation of CRAN policy.

### DIFF
--- a/R/sim.R
+++ b/R/sim.R
@@ -179,10 +179,10 @@ sim <- function (ode = NULL,
       }
     }
     if(is.null(analytical)) {
-      if("function" %in% class(ode) && is.null(attr(ode, "cpp")) || attr(ode, "cpp") == FALSE) {
+      if(inherits(ode, "function") && is.null(attr(ode, "cpp")) || attr(ode, "cpp") == FALSE) {
         stop("Sorry. Non-C++ functions are deprecated as input for ODE.")
       } else {
-        if("function" %in% class(ode)) {
+        if(inherits(ode, "function")) {
           size <- attr(ode,  "size")
         } else {
           size <- attr(get(ode),  "size")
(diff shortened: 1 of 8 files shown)


R/sim.R
```diff
@@ -292,18 +292,18 @@ sim <- function (ode = NULL,
     size <- attr(ode, "size")
   }
   if (!is.null(omega)) {
-    if(class(omega) == "matrix") {
+    if(inherits(omega, "matrix")) {
       omega_mat <- omega
     }
-    if(class(omega) %in% c("numeric", "integer")) {
+    if(inherits(omega, "numeric")) {
```
jasmineirx, line 298: it's very unlikely that omega is ever going to be an integer, but if you want to catch them both you can use `mode`:

```suggestion
    if(mode(omega) == "numeric") {
```

mode will always return a logical of length 1.

cran-comments.md
```diff
@@ -32,6 +32,24 @@ Namespace in Imports field not imported from: 'BH'
 
 Maintainer response: The BH package provides Boost C++ library headers. The BH library is not used in any of the code in /R or /src, but only in the C++ templates that comes with the package (in /inst/cpp). The C++ templates are used when creating a model using the new_ode_model() function, which is then compiled on-the-fly. If the BH package is removed from the Depends field, it won't be installed and hence models cannot be compiled and run. BH cannot be moved to the LinkingTo field.
 
+* CRAN repository db overrides:
+  X-CRAN-Comment: Archived on 2022-01-19 for policy violation.
+
+  On not remving debugging information.
```
jasmineirx, line 38:
```suggestion
  On not removing debugging information.
```

jasmineirx (approved):
lgtm. switching to using `inherits` is good practice. the review comments are addressed.

i added a couple minor suggestions.

cran-comments.md
roninsightrx: that typo was in the CRAN output, so have to keep...

roninsightrx:
add link to Github in DESCRIPTION

PR #38, opened 2022-04-19
RXR-982: Run check in correct directory

karawoo:
The PKPDsim source code is in `/src` but the latest irx-r-base image changes the working directory to something else, so I've updated the jenkinsfile here to run `devtools::check()` with the absolute path.


jasmineirx (approved):
lgtm

PR #39, opened 2022-04-22
RXR-927 allow rounding to be specified in new_covariate()

roninsightrx:
This adds a feature to perform rounding of covariate times. This is especially useful in production code when assuming a certain precision for regimens, TDMs, and covariate times.


R/new_covariate.R
```diff
@@ -114,11 +116,16 @@ new_covariate <- function(
...
   }
-  cov <- list(value = new_values,
-              times = new_times,
-              implementation = implementation,
-              unit = new_unit,
-              comments = comments)
+  if(!is.null(round_times)) {
+    new_times <- round(new_times, round_times)
+  }
+  cov <- list(
```
roninsightrx, line 122: just reformatted `cov`.

jasmineirx (approved):
lgtm!

PR #40, opened 2022-05-02
RXR-973: Use isTRUE when checking whether to build

karawoo:
Minor update to our logic for checking the `build` field.  If the `build` field is missing, `model_from_api()` would error. I don't think we necessarily want to require this field in the public versions of our models, so I'm using `isTRUE()` to avoid the error.


jasmineirx (approved):
makes sense to me!

PR #42, opened 2022-05-09
RXR-1014 fix issue with multiple observation types

roninsightrx:
As far as I can tell, the issue only surfaces when we have:
- a variable that has a basline
- observation at t=0, or observation equal to dose time
- multiple observation types

There was an indexing issue in C that caused the wrong value to be exported. Also, in R, an issue with double dosing occurred with multiple observation types and t_obs at same time of dosing.


jasmineirx (approved):
lgtm, solves issues with pkpd_heparin_delavenne

PR #44, opened 2022-06-01
RXR-1048: create pkgdown site for PKPDsim

karawoo:
Basically the same setup as for clinPK, the only difference is I didn't sort the PKPDsim functions into sections on the reference page. Unlike clinPK, PKPDsim doesn't have such clear categories of functions, so I think it's easier to just list them all alphabetically.

Screen Shot 2022-06-01 at 10 42 22 AM


roninsightrx (approved):
lgtm

PR #45, opened 2022-06-02
RXR-1052 multi obs types

roninsightrx:
Another issue with multi-obs-type models. See also previous PR here: https://github.com/InsightRX/PKPDsim/pull/42. The fix implemented there fixed one issue but created another problem for other models that wasn't caught by current tests. This PR is a slight refactor of the first solution. Also added test to test for correctness for the specific model that was broken.

### DIFF
--- a/R/parse_obs_types.R
+++ b/R/parse_obs_types.R
@@ -3,6 +3,10 @@
 #' @param obs specified observation object including at least a description of
 #'       which variable(s) are associated with a particular compartment, e.g.
 #'       `list(variable="CONC", scale="1")`.
+#' @param initial is this for the initial code block in the C++ template that
+#'       initializes the variables and compartments (`TRUE`), or for the
+#'       second code block used for the rest of the dataset?
+#'
 #' @keywords internal
 parse_obs_types <- function(obs, initial = FALSE) {
   if(length(obs$variable) == 1) {
@@ -15,7 +19,7 @@ parse_obs_types <- function(obs, initial = FALSE) {
       tmp <- c(
         tmp,
         paste0(
-          "      ", str_if, " (obs_type[i+", ifelse(initial, 0, 1), 
+          "      ", str_if, " (obs_type[i+", ifelse(initial, 0, 1),
           "]==", i, ") { obs.insert(obs.end(), ", obs$variable[i], "); } ",
           str_else, " "
         )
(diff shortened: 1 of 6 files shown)


inst/cpp/sim.cpp
```diff
@@ -123,7 +123,7 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
       ode_out tmp = sim_cpp(Aupd, t_start, t_start, step_size);
       // insert time-dependent covariates scale
       // insert scale definition for observation
-      // insert saving observations to obs object(s)
+      // insert saving initial observations to obs object(s)
```
roninsightrx, line 126: in sim.cpp template, need to split this up in generated code for the initial row (to initialize the variables) of output, and the generated code for the loop through the dataset design.

R/parse_obs_types.R
```diff
@@ -15,8 +15,8 @@ parse_obs_types <- function(obs) {
       tmp <- c(
         tmp,
         paste0(
-          "      ", str_if, " (obs_type[row]==", i, ")",
```
roninsightrx, line 18: this was the crucial bug, shouldn't have replaced `i+1` with `row`. Apparently that worked (by accident) for the tested model in the previous PR, but not for some other models.

tests/testthat/test_multi_obs.R
```diff
@@ -158,8 +158,78 @@ test_that("multi-obs with baseline and obs_time = dose_time works correctly", {
...
     tmp$CONC[tmp$obs_type == 1]
   )
-  expect_equal( 
+  expect_equal(
     tmp$y[tmp$obs_type == 4],
     tmp$ACT[tmp$obs_type == 4]
   )
 })
+
+test_that("multi-obs model simlates correctly (bug 202205)", {
```
jasmineirx, line 167:
```suggestion
test_that("multi-obs model simulates correctly (bug 202205)", {
  skip_on_cran()
```

jasmineirx (changes requested):
we are still initializing `row` and incrementing it, but not using it. we should remove this variable since it is no longer necessary.

otherwise lgtm. makes sense to split up initialization and looping to me since it allows for slightly finer control.

jasmineirx (approved):
lgtm

PR #48, opened 2022-06-10
RXR-1070: fix directory of built packages

karawoo:
I was seeing a lot of "Package not created" messages when building our models. On investigation, this was due to a change in #28 where we we stopped changing the working directory, but didn't update the rest of the logic for where to find the built tarball. I've fixed it, plus another issue that was throwing a warning.


roninsightrx (approved):
lgtm

PR #49, opened 2022-06-15
RXR-1050: Fix github actions

karawoo:
pkgdown started failing after #46, I updated to the latest version of the `setup-r` and `setup-r-dependencies` actions and that seems to have resolved it.

PR #51, opened 2022-06-17
RXR-1050: Use updated pkgdown action

karawoo:
pkgdown is still failing on master after #49. I updated my version of usethis and re-ran `use_pkgdown_github_pages()`, which now generates a slightly different version of the pkgdown github action. Hopefully this should work better; if nothing else it will run (but not deploy) on pull requests, so it should be easier to tell before merging if we'll be able to build the site or not.


jasmineirx (approved):
lgtm

PR #52, opened 2022-06-17
RXR-1081: add public models

karawoo:
Adds our public models to the `inst/models` directory. I copied the README of PKPDmodels into our Getting started vignette. 

I also recently learned that if you have a vignette that is named `pkgname.Rmd`, pkgdown will automatically put a link to it in the navbar of the page, instead of within the Articles dropdown (https://pkgdown.r-lib.org/reference/build_articles.html#get-started). I updated our getting started vignette to use this convention.


jasmineirx (approved):
lgtm! that's nice that pkgdown does that!

I'm really delighted with that package.

PR #53, opened 2022-07-05
Update version

karawoo:


karawoo:
The pkgdown failure is due to a bug in the downlit package. It looks like this package was updated twice recently, first introducing the bug, and then again with a fix to the bug. The fixed version is on CRAN but only in source form, and I think the pkgdown github action installs a binary from RSPM. I confirmed that locally the previous version fails but the new version succeeds. I'm going to go ahead and merge this and we can re-run the github actions pipeline once the binary is available.

PR #54, opened 2022-07-07
move up check for parameter vs parameter table

roninsightrx:
Needed for PKPDposterior, but also a good idea b/c we just want to check the user-specified arguments.


karawoo (changes requested):
Can we add a test for this?

karawoo (approved):
lgtm, thanks!

PR #58, opened 2022-10-05
RXR-1228 fix bug when joining regimens with no t_inf

roninsightrx:
When a drug is oral or SC administered, it will not have a t_inf specified. This PR fixes the behavior and adds a test.


karawoo (approved):
lgtm

PR #60, opened 2022-11-17
RXR-1298: add info about development population to model

karawoo:
Adds `development` as an attribute, to include info about the development population of the model.


jasmineirx (approved):
lgtm!

we might want to skip this unit test on CRAN? I recall we had some time limits and building new ODE models is slow, and this is a non-critical functionality.

PR #61, opened 2022-11-19
RXR-1298: Include development info in installed package

karawoo:
#60 included the development info in the object returned by `new_ode_model()`, but that info wasn't included in the installed package.


jasmineirx (approved):
lgtm - feels like it should be easier to add a new field 😞

PR #62, opened 2022-11-28
RXR-1298: allow empty lists in print_list()

karawoo:
This will allow us to list development info as `"development": {}` in the model json in cases where the development population isn't known.


jasmineirx (approved):
lgtm

PR #63, opened 2022-11-30
Rewrite print_list

karawoo:
`print_list()`'s quoting behavior did not work well on nested lists:

```r
dat <- list(a = list(b = "foo"), c = "bar")
print_list(dat)
#> [1] "list (\"a\" = list(b = \"foo\"), \"c\" = bar)" # needs quotes around "bar" to be valid R syntax
print_list(dat, quote = TRUE)
#> [1] "list (\"a\" = \"list(b = \"foo\")\", \"c\" = \"bar\")" # quotes around the second list( shouldn't be there
``` 

It seems like this function only really worked well on lists that were a) not nested and b) all of the same type. A more flexible approach was to rewrite it using `capture.output()` and `dput()`. I was able to install all the model packages and run all our downstream f-tests after this change.

I've implemented this as an S3 method for lists and NULL values only, on the thinking that `dput()` output may not be a perfect representation of all types of objects, and if we start using this function for other objects we probably want to test it out carefully.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: PKPDsim
 Type: Package
 Title: Tools for Performing Pharmacokinetic-Pharmacodynamic Simulations
-Version: 1.2.1
+Version: 1.2.9999
 Date: 2022-10-13
 Authors@R: c(
         person("Ron", "Keizer", email = "ron@insight-rx.com", role = c("aut", "cre")),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: PKPDsim
 Type: Package
 Title: Tools for Performing Pharmacokinetic-Pharmacodynamic Simulations
-Version: 1.2.9999
+Version: 1.2.1
 Date: 2022-10-13
 Authors@R: c(
         person("Ron", "Keizer", email = "ron@insight-rx.com", role = c("aut", "cre")),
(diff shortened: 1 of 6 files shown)


tests/testthat/test_misc.R
```diff
@@ -8,8 +8,29 @@ test_that("now_utc returns time in UTC", {
...
+  expect_equal(res1, "list(x = 1, y = 2)")
+  expect_equal(res2, "list(x = 1, y = \"foo\")")
+  expect_equal(res3, "list(x = list(y = \"foo\"), z = \"bar\")")
+})
+
+test_that("print_list can remove outer list() call", {
+  l <- list(x = 1, y = 2)
+  expect_equal(print_list(l, wrapper = FALSE), "x = 1, y = 2")
+})
+
```
jasmineirx, line 27: it works totally for for the case, but maybe one of these tests should have a vector? e.g.:

```
print_list(list(formulation = c("a", "b")), wrapper = F)
```

DESCRIPTION
```diff
@@ -1,7 +1,7 @@
 Package: PKPDsim
 Type: Package
 Title: Tools for Performing Pharmacokinetic-Pharmacodynamic Simulations
-Version: 1.2.1
+Version: 1.2.9999
```
jasmineirx, line 4: a little bit of a jump?

jasmineirx (approved):
lgtm

DESCRIPTION
karawoo: whoops, didn't mean to commit that, I just changed it so that when I installed locally I could make sure I was using the correct version. will revert.

PR #65, opened 2023-01-27
RXR-1346: add JSON to model package

roninsightrx:


R/new_ode_model.R
```diff
@@ -293,8 +297,27 @@ new_ode_model <- function (model = NULL,
...
+        overwrite = TRUE,
+        recursive = TRUE,
+        copy.mode = FALSE
+      )
+
+      ## copy definition JSON, if model created from definition file.
+      if(!is.null(definition) && !is.null(attr(definition, "path"))) {
+        dir.create(file.path(new_folder, "inst"))
+        copy_result <- file.copy(
+          from = attr(definition, "path"),
```
roninsightrx, line 312: I think it's slightly nicer to copy the original JSON file into the package, rather than the serialized & unserialized JSON. That's why I'm not just doing `jsonlite::toJSON(definition)` and saving to file. But open for other opinions.

R/read_model_json.R
```diff
@@ -7,11 +7,20 @@
 #' @md
 #' @export
 #' @return List containing contents of original JSON file
-read_model_json <- function(path) {
+read_model_json <- function(
```
roninsightrx, line 10: just added some extra protection

R/new_ode_model.R
```diff
@@ -293,8 +297,27 @@ new_ode_model <- function (model = NULL,
...
+          to = file.path(new_folder, "inst", "definition.json")
+        )
+        if(isTRUE(copy_result)) {
+          message("* Adding model definition to R package.")
+        } else {
+          message("* Could not add model definition to R package.")
+        }
+      }
```
karawoo, line 320: Maybe make this conditional on the `verbose` argument?

karawoo (commented):
Curious if there is a reason why we can't pass the `url` argument directly from `model_from_api()` to `new_ode_model()` and have `new_ode_model()` use that to copy the file? Attaching an attribute to the definition list object and passing the whole object just for `new_ode_model()` to extract the path attribute seems complicated when we're already passing the path to `model_from_api()`. But maybe I'm missing something?

roninsightrx:
> Curious if there is a reason why we can't pass the `url` argument directly from `model_from_api()` to `new_ode_model()` and have `new_ode_model()` use that to copy the file? Attaching an attribute to the definition list object and passing the whole object just for `new_ode_model()` to extract the path attribute seems complicated when we're already passing the path to `model_from_api()`. But maybe I'm missing something?

Ah yes, not sure why I missed that. Let me try.

R/new_ode_model.R
```diff
@@ -293,8 +297,27 @@ new_ode_model <- function (model = NULL,
...
+        recursive = TRUE,
+        copy.mode = FALSE
+      )
+
+      ## copy definition JSON, if model created from definition file.
+      if(!is.null(definition) && file.exists(definition)) {
+        dir.create(file.path(new_folder, "inst"))
+        copy_result <- file.copy(
+          from = definition,
+          to = file.path(new_folder, "inst", "definition.json")
```
karawoo, line 313:
```suggestion
          to = file.path(new_folder, "inst", "definition.json5")
```

karawoo (approved):
lgtm, though I think we should keep the .json5 extension. Edit: actually, if the files are not required to be json5 (and ours just happen to be json5), maybe we should just keep whatever the original extension (and possibly the original filename)?

R/new_ode_model.R
roninsightrx: Yup, if we just copy like this, it should be `.json5` since there are multiline strings that regular JSON-parsers can't handle. I'd prefer to keep `definition` though to have the same name in each package.

jasmineirx (approved):
lgtm

PR #66, opened 2023-02-23
RXR-1226: Determine fixed params from size of omega matrix if not specified

karawoo:
Adds a function called `get_fixed_parameters()` which returns the fixed parameters if specified, or determines them based on the omega matrix size if not.


roninsightrx (approved):
ok

PR #67, opened 2023-02-24
RXR-1373: prep for CRAN release

karawoo:
Per the email from CRAN, I've added `PKG_CPPFLAGS = -D_HAS_AUTO_PTR_ETC=0` to `src/Makevars`. If I understand the email correctly I think this should resolve the issue they emailed us about. However, I was not able to reproduce the issue using the [clang16 docker container](https://r-hub.github.io/containers/containers.html#clang16) from R-hub, so I am not certain that this will fix it on CRAN's end.

I also included a few minor fixes to things that caused NOTEs in R CMD check, and updated the DESCRIPTION file with a new version number and date.


roninsightrx (approved):
Agree. We'll see if something shows up during submission.

jasmineirx (approved):
lgtm

PR #68, opened 2023-02-27
Cran 1.3.0

roninsightrx:
Some final tweaks for CRAN release of 1.3.0.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -16,3 +16,4 @@ vignettes/example-pkpd-model.Rmd
 vignettes/PKPDsim.Rmd
 vignettes/variability.Rmd
 Dockerfile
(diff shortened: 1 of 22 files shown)


karawoo:
Per the [CRAN policies](https://cran.r-project.org/web/packages/policies.html) we should also probably bump the version to 1.3.1

> increasing the version number at each submission reduces confusion so is preferred even when a previous submission was not accepted.

roninsightrx:
> Per the [CRAN policies](https://cran.r-project.org/web/packages/policies.html) we should also probably bump the version to 1.3.1
> 
> > increasing the version number at each submission reduces confusion so is preferred even when a previous submission was not accepted.

Ah, didn't see this in time. CRAN accepted as 1.3.0, so let's keep that.

PR #69, opened 2023-03-06
RXR-1379: require fixed parameters in model definition

karawoo:
This reverses the changes in #66; we now require `"fixed"` to be present in the model definition.


roninsightrx (approved):
lgtm

PR #70, opened 2023-05-17
RXR-1470: circumvent difference in how NONMEM and PKPDsim handle bioavailability for infusions

roninsightrx:
PKPDsim automatically applies the bioavailability to infusions. NONMEM doesn’t do that, it instead recalculates duration as ((AMT/bioav)*rate), and then uses the given RATE to infuse. So for bioavailability of 10%, the duration will be 10 times shorter. But that's generally not really what one wants. The more intuitive way is to  just multiple RATE by bioavailability.

Generally we only have a bioavailability factor for oral models, which use a bolus into a dose compartment, so this issue will never surface. However, sometimes there is a use case for IV models too, e.g. when using it to convert unit from mg to mmol, or drug to prodrug. NONMEM's behavior with this scenario is also explained in the [RxODE manual](https://nlmixrdevelopment.github.io/RxODE-manual/events.html), as RxODE matches NONMEM's behavior.

To be clear, the behavior in PKPDsim was correct, and we’re not changing that. It’s just slightly different from NONMEM / RxODE, which IMO is less intuitive. Here, we're only updating the exporter to NONMEM datasets, so it matches behavior in PKPDsim and can be used for automated validations.

### DIFF
--- a/PKPDsim.Rproj
+++ b/PKPDsim.Rproj
@@ -17,5 +17,6 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
+PackageCleanBeforeInstall: No
 PackageInstallArgs: --no-multiarch --with-keep.source
 PackageRoxygenize: rd,collate,namespace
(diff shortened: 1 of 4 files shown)


karawoo (approved):
lgtm

PR #74, opened 2023-08-29
IRX-5176 fix variables have value at first row

roninsightrx:
When state A is initialized manually and is not [0, 0, 0, ...], we also want to have variables that depend on this initial state updated. This was not done, except for the standard observation variable `y`. We need this whenever we use a non-default concentration observation, e.g. CONC or CONCF.

### DIFF
--- a/R/compile_sim_cpp.R
+++ b/R/compile_sim_cpp.R
@@ -253,7 +253,7 @@ compile_sim_cpp <- function(
   }
   cpp_code[idx18] <- paste0(
     "  boost::array<double, ",size,"> b = { ", paste(rep(0, size), collapse=", ")," };\n",
-    "  const state_type& A_dum = b;\n",
+    "  state_type& A_dum = b;\n",
     "  state_type dAdt_dum = b;")
   sim_func <-
     paste0(paste0(readLines(paste0(folder, "/cpp/sim_header.cpp")), collapse = "\n"),
--- a/inst/cpp/sim.cpp
+++ b/inst/cpp/sim.cpp
@@ -84,11 +84,14 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
   prv_dose = doses[0];
   t_prv_dose = times[0];
 
-  // Main call to ODE solver
-  ode(A_dum, dAdt_dum, 0);
-
   // insert_state_init
   NumericVector Aupd = clone(A);
+  
+  // Main call to ODE solver
+  for(int i = 0; i < n_comp; i++) { // make sure A and variables in ode block are initialized before start
+    A_dum[i] = Aupd[i];
+  }
+  ode(A_dum, dAdt_dum, 0);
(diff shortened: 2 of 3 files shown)


R/compile_sim_cpp.R
```diff
@@ -253,7 +253,7 @@ compile_sim_cpp <- function(
   }
   cpp_code[idx18] <- paste0(
     "  boost::array<double, ",size,"> b = { ", paste(rep(0, size), collapse=", ")," };\n",
-    "  const state_type& A_dum = b;\n",
+    "  state_type& A_dum = b;\n",
```
roninsightrx, line 256: this cannot be a `const` anymore, since we're changing it to hold the initial state and not just zeroes.

inst/cpp/sim.cpp
```diff
@@ -84,11 +84,14 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
   prv_dose = doses[0];
   t_prv_dose = times[0];
 
-  // Main call to ODE solver
-  ode(A_dum, dAdt_dum, 0);
-
   // insert_state_init
   NumericVector Aupd = clone(A);
+
```
roninsightrx, line 89: this part is just re-ordering. We first need to initialize the state, and then call the ode solver once to initialize the derived variables. It was the opposite way. I don't see a reason why this change could lead to issues since there is no reasonable way the initial state should depend on something in the ODE block, it can only be the other way around.

inst/cpp/sim.cpp
```diff
@@ -117,15 +120,14 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
         start = 0;
       }
     }
-    if(start == 0) { // make a separate call to sim_cpp to make sure observation variables are initialized and stored
```
roninsightrx, line 120: This was just clean up, not related to this issue. The extra ode call was not actually needed at `start==0`, it just needs to be called at least once _before_ the variables are copied to the collection objects.

tests/testthat/test_t_init.R
```diff
@@ -3,34 +3,37 @@
...
+  n_ind = 1,
+  regimen = reg,
+  only_obs = TRUE,
+  output_include = list(variables = TRUE),
+  t_init = 10
+)
 
 test_that("TDM before first dose is considered a true initial value", {
   skip_on_cran()
-  par   <- list(CL = 7.67, V = 97.7, TDM_INIT = 500)
```
jasmineirx, line 9: the skip_on_cran() line can be removed from this code chunk. the reason it is here is because models take some time to compile and we were trying to get our total time down. this test_that block no longer takes much time to run so there is no reason to skip it, unless we also want to wrap the `new_ode_model` model and related lines such that they run only if not on cran.

jasmineirx (approved):
lgtm, one comment about the unit tests.

PR #77, opened 2023-10-04
use LOCF when non-numeric too; check for implementation type

jasmineirx:
There was a unique situation that was not being handled correctly: 

- if the value supplied was non-numeric and 
- the `remove_negative_times` argument was `TRUE` (default), and 
- there were negative times, with the non-numeric value changing on either side of `t = 0`

then we were not actually carrying the last observation forward, but instead filtering the observations.

This is a pretty niche situation (`PKPDsim::sim`, the intended consumer of this data object requires numeric values) but when these values are used as placeholders for later use it was leading to some confusing results.

For safety, I also added a check to ensure the implementation method specified matches what we say in the documentation are the acceptable input types.

### DIFF
--- a/R/new_covariate.R
+++ b/R/new_covariate.R
@@ -30,6 +30,7 @@ new_covariate <- function(
     times <- c(0)
     value <- value[1]
   }
+  implementation <- match.arg(tolower(implementation), c("interpolate", "locf"))
   srt <- order(times)
   times <- times[srt]
   values <- value[srt]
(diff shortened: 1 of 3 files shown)


R/new_covariate.R
```diff
@@ -70,21 +71,19 @@ new_covariate <- function(
       if(! (0 %in% new_times)) {
         # add zero time, and remove all t < 0
```
jasmineirx, line 72: the section below is just changing indentation. now the `else` logic is processed for all non-numeric information. i.e., LOCF is assumed for removing observations prior to t=0 for non numeric information. 

I thought about adding a check to ensure that `implementation == "LOCF"` for non-numeric data since interpolation does not make sense in that situation, but that's perhaps a slightly larger change to the interface. What do you think?

R/new_covariate.R
```diff
@@ -30,6 +30,7 @@ new_covariate <- function(
     times <- c(0)
     value <- value[1]
   }
+  implementation <- match.arg(tolower(implementation), c("interpolate", "locf"))
```
karawoo, line 33: Maybe we should make the default value for the `implementation` argument `c("interpolate", "locf")` to convey what the two options are, then this line could just be `implementation <- match.arg(implementation)`?

R/new_covariate.R
```diff
@@ -70,21 +71,19 @@ new_covariate <- function(
...
-              t2 <- utils::head(new_times[new_times > 0],1)
-              grad <-  (y2-y1) / (t2-t1)
-              new_values <- c(new_values, y1 + grad * (0-t1))
-              new_unit <- c(new_unit, utils::tail(new_unit[new_times < 0],1))
-            } else { # add the last obs before 0 as t=0
-              new_times <- c(new_times, 0)
-              new_values <- c(new_values, utils::tail(new_values[new_times < 0], 1))
-              new_unit <- c(new_unit, utils::tail(new_unit[new_times < 0], 1))
-            }
+          if(mode(new_values) == "numeric" && implementation == "interpolate") {
```
karawoo, line 74: Earlier in the code at line 43 we have this logical condition:
```
if(implementation == "interpolate" && inherits(values, "numeric") && !is.null(interpolation_join_limit) && interpolation_join_limit > 0)
```

`mode(new_values) == "numeric"` will return `TRUE` for both doubles and integers; `inherits(values, "numeric")` will return `FALSE` for integers. Should we use `mode(values) == "numeric"` up at line 43 too for consistency?

R/new_covariate.R
karawoo: I do think it's worth erroring or at least warning if there is a certain combination of arguments that isn't valid (and maybe putting that in the documentation too).

R/new_covariate.R
jasmineirx: I've updated as per your suggestion. Since we use a mix of 'locf' and 'LOCF' across documentation and testing I wanted to make sure we were not case-sensitive in our check so it was slightly more complicated.

R/new_covariate.R
jasmineirx: I've updated accordingly. Since we now also switch to LOCF if the values are non-numeric, this check was also greatly simplified.

karawoo (approved):

roninsightrx (approved):
lgtm

PR #80, opened 2023-11-22
RXR-1771: Remove constraints on regimen type

karawoo:
Removes logic that forces `type` to be one of 'bolus', infusion', 'oral', 'sc' or 'im'. If `type` is `NULL` we will still assume it's infusion if there's a `t_inf`, bolus otherwise.


jasmineirx (approved):
lgtm!

PR #81, opened 2023-12-07
rm assumption non-inf types are t_inf = o

jasmineirx:
In create_event_table, two changes were necessary to enable simultaneous infusion-type doses in different compartments.

1) needed to assume that infusion doses could have a "type" other than "infusion". Changed logic to reference `t_inf`, assuming the user specified this correctly as non-zero.
2) needed to fix logic to also consider dose cmt when looking for duplicated rows

Added a test too

### DIFF
--- a/R/create_event_table.R
+++ b/R/create_event_table.R
@@ -119,7 +119,7 @@ create_event_table <- function(
   regimen$t_inf[is_bolus] <- 0
   dos <- data.frame(cbind(t = regimen$dose_times,
                   dose = regimen$dose_amts,
-                  type = as.numeric(!is_bolus),
+                  type = as.numeric(regimen$t_inf != 0),
                   dum = 0,
                   dose_cmt = regimen$dose_cmt,
                   t_inf = regimen$t_inf,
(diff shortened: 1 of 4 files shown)


karawoo:
@jasmineirx can you look into the failing tests before I review?

jasmineirx:
Thanks, the tests revealed an error I hadn't noticed where "dummy" covariate rows were being given a `type` of "0". Previously we were checking for if the `type` was "infusion" to go through the "should drugs have an infusion rate" logic, and I had changed it to use "is t_inf > 0", and changed the `type` numerical conversion to instead look for bolus/oral -> 0 and everything else -> 1. I've updated it so that `type` consistently uses t_inf to determine if that row is type=1 (has a rate) or type=2 (no rate). Now covariate rows also have a type of 0, as before.

jasmineirx:
if you would like to try out the test with a model with multiple dosing compartments, you can try this model: https://github.com/InsightRX/models/pull/598

karawoo (approved):
thanks, lgtm, I tested with that model and it worked as expected

PR #82, opened 2023-12-14
RXR-1771: Include sc, im among types whose t_inf should be 0

karawoo:
We were getting errors for a regimen that was a mix of bolus and sc doses. `t_inf` for the boluses was being set to 0, but for sc it was `NA`. This led to an error at line 129. I made two changes: added sc and im to the vector of types whose infusion length should be set to 0. Also added `na.rm = TRUE` at line 129 for greater safety.

### DIFF
--- a/R/create_event_table.R
+++ b/R/create_event_table.R
@@ -113,9 +113,9 @@ create_event_table <- function(
   }
 
   # parse list to a design (data.frame)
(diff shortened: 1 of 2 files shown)


jasmineirx (approved):

roninsightrx (approved):
Lgtm

PR #83, opened 2023-12-20
RXR-1802: fix issue with parsing of a long line

roninsightrx:
when compilation of a model with a 5 different output variables, the parsing when wrong because the parser split it into multipel lines, which messed up the template `model.R` file in the generated package.

I don't think a test is necessary, simple fix.


jasmineirx (approved):
LGTM, fixes the issue I was having with a model that invovled 6 types of samples.

PR #84, opened 2023-12-20
busulfan models: mccune and shukla

jasmineirx:
This PR adds two new models to our open source model library: the McCune model and the Shukla model. Both models have been previously published (citation in model files) and were used as part of a comparison of NCA and MAP approaches to AUC estimation presented at IATDMCT 2022 (Prague).


roninsightrx (approved):
lgtm

PR #86, opened 2024-01-06
Fix `n` and length of rate vector in regimens joined by t_dose_update

karawoo:
When joining regimens using `t_dose_update`, only the `rate` values from the first regimen were included, and the value in `n` was incorrect. This fixes both.

### DIFF
--- a/R/join_regimen.R
+++ b/R/join_regimen.R
@@ -45,6 +45,8 @@ join_regimen <- function(
     regimen1$t_inf <- c(regimen1$t_inf[keep], regimen2$t_inf)
     regimen1$interval <- regimen2$interval
     regimen1$type <- c(regimen1$type[keep], regimen2$type)
+    regimen1$rate <- c(regimen1$rate[keep], regimen2$rate)
+    regimen1$n <- length(regimen1$dose_amts)
     return(regimen1)
   }
 
--- a/tests/testthat/test_join_regimen.R
+++ b/tests/testthat/test_join_regimen.R
@@ -74,3 +74,20 @@ test_that("join regimen works when no t_inf specified, e.g. oral or sc dosing", 
   expect_equal(reg$type, c("sc", "sc", "sc", "sc", "sc", "sc", "sc", "sc"))
   expect_equal(reg$dose_amts, c(500, 500, 500, 400, 400, 400, 400, 400))
 })
+
+test_that("join_regimen with t_dose_update maintains correct rate length", {
+  reg1 <- new_regimen(
+    amt = 1,
+    times = c(0, 24),
+    type = "oral"
+  )
+  reg2 <- new_regimen(
+    amt = 2,
+    time = c(0, 24, 48),
+    type = "oral"
+  )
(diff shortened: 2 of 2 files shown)


jasmineirx (changes requested):
should we also update the $rate and $n fields for all the other joins in this function?

karawoo:
I thought this was already working for the other branches of the logic because the following works:

``` r
library("PKPDsim")
reg1 <- new_regimen(
  amt = 1,
  times = c(0, 24),
  type = "oral"
)
reg2 <- new_regimen(
  amt = 2,
  time = c(0, 24, 48),
  type = "oral"
)
reg_b <- join_regimen(reg1, reg2, dose_update = 2)
str(reg_b)
#> List of 8
#>  $ interval       : num 24
#>  $ n              : int 4
#>  $ type           : chr [1:4] "oral" "oral" "oral" "oral"
#>  $ t_inf          : num [1:4] 0 0 0 0
#>  $ dose_times     : num [1:4] 0 24 48 72
#>  $ dose_amts      : num [1:4] 1 2 2 2
#>  $ rate           : num [1:4] 0 0 0 0
#>  $ first_dose_time: POSIXct[1:1], format: "2024-01-08 18:05:03"
#>  - attr(*, "class")= chr [1:2] "regimen" "list"
```

<sup>Created on 2024-01-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

But actually, it only works by filling in the default rate of 0 because these are oral regimens. However when I tried to update the logic, I realized that `new_regimen()` actually does not return `rate` at all for infusion regimens:
``` r
library("PKPDsim")
str(new_regimen(amt = 1, times = 0, type = "infusion", rate = 0.5))
#> List of 7
#>  $ interval       : num 24
#>  $ n              : int 1
#>  $ type           : chr "infusion"
#>  $ t_inf          : num 2
#>  $ dose_times     : num 0
#>  $ dose_amts      : num 1
#>  $ first_dose_time: POSIXct[1:1], format: "2024-01-08 18:04:32"
#>  - attr(*, "class")= chr [1:2] "regimen" "list"
```

<sup>Created on 2024-01-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

So to fix `join_regimen()` we'd need to also update `new_regimen()` so that `rate` is included for infusions. I can do this, but I wonder if it would instead make more sense to remove `rate` from non-infusion regimens instead? It seems a bit odd that we are currently including it only in regimens where it is not relevant.

jasmineirx:
When we simulate using the regimens, if there are non-zero infusion rates we calculate rate in `create_event_table` and otherwise we set rate to zero

https://github.com/InsightRX/PKPDsim/blob/master/R/create_event_table.R#L120-L131

so it's not clear to me how we are using rate -- possibly in other checks or times we extract info from regimens. I am ok with merging for now and tackling separately making rate work more consistently.

jasmineirx (approved):

karawoo:
Sounds good, I added https://insight-rx.atlassian.net/browse/RXR-1833 to address the rate issue more generally

PR #87, opened 2024-01-10
fix regimen to nm convertor

roninsightrx:


jasmineirx (approved):
lgtm. changed logic slightly to be consistent with how we have handled this elsewhere.

PR #88, opened 2024-01-12
ensure if always has length 1 result

jasmineirx:
One more issue arose with the updates to regimen_to_nm in which we had a condition length > 1 if we allowed for multiple dose compartments within one regimen (since the `if` statement condition would be of length > 1).

I accidentally also committed some changes I was making to get_t_obs_from_regimen to update the code base to no longer assume all infusion regimens were of type "infusion". I think it's okay to bring them in too.


roninsightrx (approved):
ok!

PR #89, opened 2024-02-12
RXR-1884: timevarying covariates were not interpolated when in PK block

roninsightrx:
Time-varying covariates were only really interpolated when they were included in the `ode` block, and not when in the `pk` block. E.g. compare:

Not working previously:
```
  code = "
      dAdt[1] = -(Q/V)*A[1] + (Q/V2)*A[2] -(CLi/V)*A[1];
      dAdt[2] = -(Q/V2)*A[2] + (Q/V)*A[1];
    ",
  pk_code = "CLi = CL + CRCL",
```

Working fine:
```
  code = "
     CLi = CL + CRCL
      dAdt[1] = -(Q/V)*A[1] + (Q/V2)*A[2] -(CLi/V)*A[1];
      dAdt[2] = -(Q/V2)*A[2] + (Q/V)*A[1];
    "
```

In the output from `sim()` it looked as if covariates were actually interpolated over time, but the interpolated value was not actually applied in the PK calculations. The covariates were still changing over time, but always using the value of the last change (similar to `locf` method).

This PR fixes the behavior, i.e. it executes the code that calculates the interpolated covariates at the right place, i.e. just before the PK block is executed, and not afterwards.

Impact: overall, the impact is most likely minimal for most models. It is only for very strong covariates and/or covariates that are changing rapidly over time where this impacts simulations. It also only impacts models where the code was in the `pk` block.

Note: PK code is evaluated once every event (dose start/stop, observation time, or covariate change). In principle, for most purposes it is fine to write any PK code directly in the ODE block. Only when code needs to be performant, it makes sense to write the code in the PK block instead.

### DIFF
--- a/R/compile_sim_cpp.R
+++ b/R/compile_sim_cpp.R
@@ -142,7 +142,7 @@ compile_sim_cpp <- function(
   cpp_code[idx] <- par_def
   idx2 <- grep("insert_state_init", cpp_code)
   cpp_code[idx2] <- paste("   ", code_init)
-  cov_scale <- ""
+  cov_timevar <- ""
   cov_names <- NULL
   if(!is.null(covariates)) {
     if(inherits(covariates, "character")) {
@@ -176,7 +176,7 @@ compile_sim_cpp <- function(
         cov_tmp <- paste0(cov_tmp, paste0('    t_prv_', nam, ' = 0 ;\n'))
       }
       cov_tmp <- paste0(cov_tmp, paste0('    ', nam, ' = ', nam,'_0;\n'))
-      cov_scale <- paste0(cov_scale, paste0('      ', nam, ' = ', nam, '_0 + gr_', nam, ' * (tmp.time[k] - t_prv_', nam), ');\n')
+      cov_timevar <- paste0(cov_timevar, paste0('      ', nam, ' = ', nam, '_0 + gr_', nam, ' * (t_end - t_prv_', nam), ');\n')
     }
     idx3 <- grep("insert covariate definitions", cpp_code)
     cpp_code[idx3] <- cov_def
@@ -208,7 +208,7 @@ compile_sim_cpp <- function(
     if(length(obs$cmt) == 1) {
       cpp_code[idx5] <- paste0(cpp_code[idx5], "\n    scale = ", obs$scale[1], ";")
       cpp_code[idx7] <- paste0(cpp_code[idx7], "\n      scale = ", obs$scale[1], ";")
-      cpp_code[idx8] <- cov_scale
+      cpp_code[idx8] <- cov_timevar
       cpp_code[idx12a] <- paste0(cpp_code[idx12a], "\n      obs.insert(obs.end(), tmp.y[k][", obs$cmt[1]-1,"] / scale);")
       cpp_code[idx12b] <- paste0(cpp_code[idx12b], "\n      obs.insert(obs.end(), tmp.y[k][", obs$cmt[1]-1,"] / scale);")
       cpp_code[idx13] <- paste0('  comb["obs"] = obs;\n');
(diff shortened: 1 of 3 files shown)


R/compile_sim_cpp.R
```diff
@@ -142,7 +142,7 @@ compile_sim_cpp <- function(
   cpp_code[idx] <- par_def
   idx2 <- grep("insert_state_init", cpp_code)
   cpp_code[idx2] <- paste("   ", code_init)
-  cov_scale <- ""
+  cov_timevar <- ""
```
roninsightrx, line 145: just a name change so it's clearer what this does.

R/compile_sim_cpp.R
```diff
@@ -176,7 +176,7 @@ compile_sim_cpp <- function(
         cov_tmp <- paste0(cov_tmp, paste0('    t_prv_', nam, ' = 0 ;\n'))
       }
       cov_tmp <- paste0(cov_tmp, paste0('    ', nam, ' = ', nam,'_0;\n'))
-      cov_scale <- paste0(cov_scale, paste0('      ', nam, ' = ', nam, '_0 + gr_', nam, ' * (tmp.time[k] - t_prv_', nam), ');\n')
+      cov_timevar <- paste0(cov_timevar, paste0('      ', nam, ' = ', nam, '_0 + gr_', nam, ' * (t_end - t_prv_', nam), ');\n')
```
roninsightrx, line 179: the gradient calculation is based on the time difference between the last change in the covariate value, and the end of the current integration period.

inst/cpp/sim.cpp
```diff
@@ -106,6 +106,7 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
       t_prv_dose = times[i];
       prv_dose = doses[i];
     }
+    // insert time-dependent covariates scale
```
roninsightrx, line 109: this is the label that `compile_sim_cpp` is looking for to replace with code. This needed to be moved up, to be executed before the call to `pk_code()`

inst/cpp/sim.cpp
```diff
@@ -126,7 +127,6 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
     ode_out tmp = sim_cpp(Aupd, t_start, t_end, step_size);
     if(start == 0) { // make sure observation variables are stored
       int k = 0;
-      // insert time-dependent covariates scale
```
roninsightrx, line 129: this is not needed. It was calculating the covariate values in the output table. No need to recompute.

tests/testthat/test_timevar_cov.R
```diff
@@ -1,26 +1,28 @@
...
   # a new steady state that is much lower
-
   covs <- list(
     CRCL = new_covariate(
       value = c(4.662744, 5.798767, 6.195943, 6.2, 600),
       times = c(0, 18, 29, 209, 210),
       implementation = "locf"
     )
   )
-  mod <- new_ode_model(
```
roninsightrx, line 14: moved this out as it can be shared between the two tests

jasmineirx (approved):
lgtm. tested that it also worked in nova.

in order to assess the impact on f-tests, we need to re-install all the models that are used in f-tests. my hunch is that most f-tests we create have limited covariates per treatment course, but there might be a few based on Gold Copies that do have some covariate variation.

roninsightrx:
> lgtm. tested that it also worked in nova.
> 
> in order to assess the impact on f-tests, we need to re-install all the models that are used in f-tests. my hunch is that most f-tests we create have limited covariates per treatment course, but there might be a few based on Gold Copies that do have some covariate variation.

Yup, I ran those and somewhat surprisingly only one test failed. I'm looking into that one to see if it perhaps has a different cause, or if we just need to update the reference value. 

I agree many f-tests will not actually test the timevarying covariates, but I expect that we have at least two dozen or so where we do have them. But we have to realize that for this issue to make a meaningful difference you have to have a large difference in covariate value _and_ a large difference in timing of those samples (while continuing dosing). Most common time-varying covariate like creats are usually taken every day or every other day, and probably don't jump enough to make a difference.

roninsightrx:
> I'm looking into that one to see if it perhaps has a different cause, or if we just need to update the reference value.

Even that single one was a false positive. So we had zero tests that caught this. Which is good (in that it apparently doesn't matter much in practice) and bad (that we should have at least some that do capture differences, even if minute). Will add an f-test.

PR #91, opened 2024-03-06
RXR-1851: add function to calculate auc pre-ss

roninsightrx:
This adds a convenience function to calculate AUC (and conc) using linear models, based on just the PK parameters.


R/calc_auc_analytic.R
```diff
@@ -0,0 +1,88 @@
...
+    t_obs = c(0, 24, 48, 72),
+    ...
+) {
+  model <- match.arg(f)
+
+  if(is.null(regimen)) {
+    if(is.null(dose) || is.null(interval)) {
+      stop("Specify `regimen` or `dose` and `interval`")
+    }
+    if(is.null(t_inf) || t_inf == 0) t_inf <- 1e-6
```
jasmineirx, line 49:
```suggestion
    if(is.null(t_inf) || any(t_inf == 0)) t_inf[t_inf > 0] <- 1e-6
```

t_inf can be of length > 1. the suggested code handles this situation.

jasmineirx (approved):
lgtm

one consideration is that this requires standard mapping of doses into the first compartment, and so won't support drugs like tacrolimus where we often have a mix of oral/iv dosing. do you think it makes sense to add some documentation or warnings somewhere? one option could be to check `reg` for multiple `type` arguments. the user is more likely to be aware of the limitation if supplying dose/interval/t_inf since the "type" argument is not present.

tests/testthat/test_calc_auc_analytic.R
```diff
@@ -0,0 +1,114 @@
...
+  res <- sim(
+    mod,
+    parameters = par,
+    regimen = reg,
+    covariates = covs,
+    t_obs = t_obs
+  ) %>%
+    dplyr::filter(comp == attr(mod, "size")) %>%
+    dplyr::mutate(auc = round(c(0, diff(y)), 3)) %>%
+    select(t, auc)
```
jasmineirx, line 23:
```suggestion
  res <- sim(
    mod,
    parameters = par,
    regimen = reg,
    covariates = covs,
    t_obs = t_obs
  ) %>%
  res <- res[res$comp == attr(mod, "size"))
  res$auc <- round(c(0, diff(res$y)), 3)
  res <- res[, c("t", "auc")]
```

This is causing an issue with R CMD check.

roninsightrx:
> lgtm
> 
> one consideration is that this requires standard mapping of doses into the first compartment, and so won't support drugs like tacrolimus where we often have a mix of oral/iv dosing. do you think it makes sense to add some documentation or warnings somewhere? one option could be to check `reg` for multiple `type` arguments. the user is more likely to be aware of the limitation if supplying dose/interval/t_inf since the "type" argument is not present.

> Right now it can only be used for iv models, so it shouldn't be an issue. The linear oral models in PKPDsim don't calculate AUC.

jasmineirx (approved):
lgtm

PR #92, opened 2024-03-08
RXR-1926 implement infusion duration scaling

roninsightrx:
This implements a feature where a model automatically scales infusion durations by a specified factor.
An example model where this is used in NONMEM is Centanni et al. CPK 2024.

In the implementation in PKPDsim it is handled in the same was as lagtime, it just modifies the input regimen / event table that is used in the simulation function.


R/apply_duration_scale.R
```diff
@@ -0,0 +1,54 @@
...
+#' @param cmt_mapping map of administration types to compartments, e.g. `list("oral" = 1, "infusion" = 2, "bolus" = 2)`.
+#'
+#' @export
+#'
+#' @return Original regimen with infusion lengths scaled by a factor
+#'
+apply_duration_scale <- function(
+  regimen,
+  duration_scale = NULL,
+  parameters,
```
jasmineirx, line 22: from the documentation, it seems like parameters are only required in certain situations ("if parameters are specified"). should this therefore default to NULL? I think the documentation could be a little clearer about what this argument is and when it is necessary

R/apply_duration_scale.R
```diff
@@ -0,0 +1,54 @@
...
+  }
+  if(!is.null(duration_scale)) {
+    if(class(duration_scale) %in% c("numeric", "integer")) {
+      if(length(duration_scale) == 1) {
+        regimen$t_inf <- regimen$t_inf * duration_scale
+      } else {
+        regimen$t_inf <- regimen$t_inf * duration_scale[regimen$cmt]
+      }
+    }
+    if(class(duration_scale) %in% c("character")) {
```
jasmineirx, line 37:
```suggestion
    } else if (class(duration_scale) %in% c("character")) {
```

slightly clearer, since you never want to do both conversions

R/apply_duration_scale.R
```diff
@@ -0,0 +1,54 @@
...
+      if(length(duration_scale) == 1) {
+        regimen$t_inf <- regimen$t_inf * parameters[[duration_scale]]
+      } else {
+        if(is.null(regimen$cmt)) {
+          if(!is.null(cmt_mapping)) {
+            regimen$cmt <- as.numeric(cmt_mapping[regimen$type])
+          } else {
+            regimen$cmt <- rep(1, length(regimen$dose_times))
+          }
+        }
```
jasmineirx, line 47: should we pull this out and put it at the top of the function in case the logic that converts a numeric `duration_scale` also needs it?

jasmineirx (changes requested):
a couple design question. Can you also please add some unit tests to confirm the function is working as intended?

R/apply_duration_scale.R
roninsightrx: Hmm, i copied from `apply_lagtime.R`. Gonna check if this is really needed in both functions. If so, it indeed looks like it will make more sense at the top.

roninsightrx: Not needed in this function, but gonna leave those lines in `apply_lagtime`, tests fail when removed.

R/sim.R
```diff
@@ -148,6 +148,14 @@ sim <- function (ode = NULL,
   if(!is.null(lagtime)) {
     regimen <- apply_lagtime(regimen, lagtime, parameters, attr(ode, "cmt_mapping"))
   }
+  if(!is.null(attr(ode, "dose")$duration_scale)) {
+    regimen <- apply_duration_scale(
+      regimen,
+      attr(ode, "dose")$duration_scale,
+      parameters,
+      attr(ode, "cmt_mapping")
```
jasmineirx, line 156:
```suggestion
      parameters
```
this argument is no longer necessary

R/apply_duration_scale.R
```diff
@@ -0,0 +1,47 @@
...
+  if(is.null(regimen$t_inf) || all(regimen$t_inf == 0)) {
+    warning("`duration_scale` is only relevant for infusion regimens with an infusion length > 0.")
+    return(regimen)
+  }
+  if(!is.null(duration_scale)) {
+    if(class(duration_scale) %in% c("numeric", "integer")) {
+      if(length(duration_scale) == 1) {
+        regimen$t_inf <- regimen$t_inf * duration_scale
+      } else {
+        regimen$t_inf <- regimen$t_inf * duration_scale[regimen$cmt]
```
jasmineirx, line 33:
```suggestion
        regimen$t_inf <- regimen$t_inf * duration_scale[regimen$type]
```

does it make more sense to use the `type` argument since i think that is how we tend to differentiate doses of different types (which go to different compartments)

jasmineirx (commented):
overall lgtm, but one question about cmt vs type

R/apply_duration_scale.R
roninsightrx: Sorry, I misread my earlier code. That 5 lines of code I removed is actually useful, at least when we're using a vector of duration scales that matches `cmt`s, which we do. The removed code ensures there is always a `cmt` in the regimen, even when not specified by the user. So will bring that code back. I think it is OK where it is, and doesn't need to go on top of the function, since we only want to make a modification to the regimen if its actually needed, IMO.

Regarding your 2nd question whether we should link it to `type` rather than `cmt`: I think it's less consistent to link it to `type`: we now have "lagtime" also linked to a specific `cmt`, not to an observation type. I see those as two very similar features. Also in NONMEM durations and infusion lengths are linked to compartments, so for most users more intuitive. An option for a future ticket could be that `lagtime` and `duration_scale` could also be specified as a list, where they then indeed can be mapped to observation types.

jasmineirx (approved):

PR #93, opened 2024-03-25
RXR-1953 modify function to calculate parameters to allow computation over time

roninsightrx:
The convenience function calculate_parameters() is intended to quickly calculate the effective parameters for a given model and covariates. However, some models have time-varying parameter definitions. To allow computations of effective parameters over time, I've now added a `t_obs` argument, which then returns each parameter as a vector of values instead of a single value.


jasmineirx (approved):
lgtm. very clearly written tests.

PR #94, opened 2024-04-03
add tests, plus to multiplication

jasmineirx:
I was trying to follow the logic in this function, since there are a lot of different combinations of arguments accepted. I added a few tests as examples of ways to supply these arguments, increasing the test coverage of the file to 100%.

I found what I think is an error: an addition should have been changed to a multiplication.

### DIFF
--- a/R/apply_duration_scale.R
+++ b/R/apply_duration_scale.R
@@ -47,7 +47,7 @@ apply_duration_scale <- function(
         regimen$t_inf <- regimen$t_inf * parameters[[duration_scale]]
       } else {
         par_tmp <- parameters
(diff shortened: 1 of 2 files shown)


roninsightrx:
Oh wow, completely missed that. Thx!

roninsightrx:
I just pushed up some changes, seems there is a conflict now.

roninsightrx (approved):
lgtm

jasmineirx:
I feel like theres a clever way to create a sub-function that handles all the logic for both apply_lagtime and apply_duration_scale, except passing either `*` or `+` as an argument, but the functions are structured just *slightly* different that it would take a bit of time.

PR #95, opened 2024-04-04
RXR-1876 rewrite function + update tests

roninsightrx:
Updates logic in regimen_to_nm() to work properly with vectors of bioavailability specifications.


roninsightrx:
Proper fix implemented [here](https://github.com/InsightRX/PKPDsim/pull/105)

PR #96, opened 2024-04-19
set up actions

roninsightrx:
I think we should leave the `--as-cran` flag, as it seems to be checking extra stuff. The `skip_on_cran()` function is not affected by the flag, I think it only looks whether it is actually running on a CRAN server or not.


jasmineirx (approved):
lgtm 🧹

PR #97, opened 2024-06-03
add t_inf to sc doses in new_regimen

JordanBrooks33:
Added t_inf argument to new_regimen() and adapted tests.

### DIFF
--- a/R/new_regimen.R
+++ b/R/new_regimen.R
@@ -68,30 +68,16 @@ new_regimen <- function(
     if (any(reg$type == "covariate")) {
       stop("'covariate' is a protected type and cannot be used for doses.")
     }
-    if(any(type == "infusion") && (is.null(t_inf) || length(t_inf) == 0)) {
-      reg$t_inf[reg$type == "infusion"] <- 1
-    } else if (any(is.na(t_inf))) {
-      reg$t_inf[(reg$type=="infusion" & is.na(t_inf))] <- 1
+    if (is.null(t_inf) || length(t_inf) == 0) {
+      reg$t_inf <- rep(NA_real_, length(reg$type))
     }
-    if(any(type == "sc") && (is.null(t_inf) || length(t_inf) == 0)) {
-      reg$t_inf[reg$type=="sc"] = 1/60
-    } else if (any(is.na(t_inf))) {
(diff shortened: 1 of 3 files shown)


JordanBrooks33:
@jasmineirx Thank you! That makes a lot of sense. I have updated the logic to assign the correct t_inf default per type. I also included im alongside sc as a default of 1/60, given the similarity of these routes. I also updated the tests you had added to be slightly broader, but still testing the same scenarios.

R/new_regimen.R
```diff
@@ -69,9 +69,29 @@ new_regimen <- function(
...
+    }
+    if(any(type == "bolus") && (is.null(t_inf) || length(t_inf) == 0)) {
+      reg$t_inf[reg$type=="bolus"] = 0
+    } else if (any(is.na(t_inf))) {
+      reg$t_inf[(reg$type=="bolus" & is.na(reg$t_inf))] <- 0
+    }
+    if(any(type == "oral") && (is.null(t_inf) || length(t_inf) == 0)) {
+      reg$t_inf[reg$type=="oral"] = 0
+    } else if (any(is.na(t_inf))) {
+      reg$t_inf[(reg$type=="oral" & is.na(reg$t_inf))] <- 0
```
jasmineirx, line 94: the blocks above are pretty repetitive.

what about something like

```r

if (is.null(t_inf) || length(t_inf) == 0) {
  reg$t_inf <- rep(NA_real_, length(reg$type))
}
if (any(is.na(reg$t_inf))) {
  # impute using standard values for each dose type
  impute_t_inf <- c(infusion = 1, sc = 1/60, im = 1/60, oral = 0, bolus = 0)
  idx <- which(is.na(reg$t_inf))
  reg$t_inf[idx] <- impute_t_inf[reg$type[idx]]
  # set unrecognized types to 1
  reg$t_inf[is.na(reg$t_inf)] <- 1
}
```

jasmineirx (commented):
I have a suggestion for reducing the repetitiveness of the code while balancing readability

R/new_regimen.R
JordanBrooks33: very nice.

jasmineirx (approved):
lgtm!

roninsightrx (approved):
lgtm

PR #98, opened 2024-06-07
RXR-2112: facilitate installation of example models

jasmineirx:
This PR adds a new helper function to make it easy to install the provided models.

You can see which models are available using:

```r
available_default_literature_models()
```

and you can install them (provided PKPDsim is installed) using

```r
install_default_literature_model("pk_busulfan_mccune")
```

or you can install all provided models:
```r
install_default_literature_model("all")
```


R/model_from_api.R
```diff
@@ -97,3 +97,48 @@ model_from_api <- function(url,
...
+#' [available_default_literature_models()]
+#' @param ... arguments passed onto `model_from_api`. For fine-grain control, it
+#'   is better to install models directly from [model_from_api()] or
+#'   [new_ode_model()].
+#' @examples \dontrun{
+#' install_default_literature_model("pk_busulfan_mccune")
+#' }
+#' @export
+
+install_default_literature_model <- function(model, ...) {
```
roninsightrx, line 116: I would probably separate into its own file? I see that it is sort of a helper function for model_from_api, but it's an exported one and kind of separate functionality?

roninsightrx (approved):
nice QoL improvement. 1 consideration, but lgtm

PR #99, opened 2024-06-10
helper functions for working with PKPDsim models

jasmineirx:
A lot of information about a model is hidden within its attributes, which are cumbersome to access via `attr` or `attributes`.

By wrapping `attr` calls for typical fields we access into functions, this information is made more accessible and better documented.


R/get_model_info.R
```diff
@@ -0,0 +1,62 @@
...
+#' Get model linearity
+#' @rdname get_model_info
+#' @returns get_model_linearity: returns a single string indicating model
+#'   linearity. E.g., "linear" or "nonlinear".
+#' @export
+get_model_linearity <- function(model) attr(model, "misc")[["linearity"]]
+
+#' Get AUC compartment
+#' @rdname get_model_info
+#' @returns get_model_auc_compartment: returns the index of the final
```
roninsightrx, line 49: yeah, at some point we should fix that and make a dedicated `auc` option or so. We shouldn't rely on this assumption. Fine for now.

R/get_model_info.R
```diff
@@ -0,0 +1,62 @@
...
+#'   variability.
+#' @export
+get_model_fixed_parameters <- function(model) attr(model, "fixed")
+
+#' Get model structure
+#' @rdname get_model_info
+#' @returns get_model_structure: returns a single string indicating model
+#'   structure. E.g.,: "1cmt_iv", "2cmt_oral".
+#' @export
+get_model_structure <- function(model) attr(model, "misc")[["model_type"]]
```
roninsightrx, line 38: should we keep same naming as the original attribute, i.e. `get_model_type` ?

roninsightrx (approved):
lgtm, 1 consideration and 1 comment.

R/get_model_info.R
jasmineirx: i considered it but it seemed so generic. i thought "structure" was clearer, since "linearity" could also be considered a type.

dominic-irx (approved):
lgtm

PR #100, opened 2024-06-10
add informative error message to facilitate debugging

jasmineirx:
I was running a simulation and encountered an uninformative error message. I was able to track down the source but it took me a while since the error was very generic:

>Error in vapply(regimen$type, function(x) cmt_mapping[[x]], FUN.VALUE = numeric(1),  : 
  values must be length 1,
 but FUN(X[[1]]) result is length 0

Having this extra check and more informative error message would have saved me some time.


dominic-irx (approved):
lgtm, tricky!

roninsightrx (approved):
👍🏻

PR #101, opened 2024-06-12
ensure paths with spaces are parsed properly

roninsightrx:
The command
```
PKPDsim::install_default_literature_model("pk_busulfan_mccune")
```
wasn't working because in `new_ode_model()` it tried to run a command like
```
R CMD INSTALL --arg1 --arg2 C:\Users\User name\etc
```
So the space in the path actually split that argument into two arguments, and hence made the installation fail. The fix just adds double-quotes around that argument to ensure it is parsed as a single argument. After this, the package installed fine. 

Hopefully this fixes the issue (https://github.com/InsightRX/mipdtrial/actions/runs/9473553934/job/26101414203) with mipdtrial, although from the error there it doesn't seem like that path actually contained a space...

The change shouldn't affect behavior on Linux or Mac, in fact it is safer there too.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -34,9 +34,6 @@ jobs:
 
       - uses: r-lib/actions/setup-pandoc@v2
       
-      - name: Install curl
-        run: sudo apt-get install -y libcurl4-openssl-dev
-
       - uses: r-lib/actions/setup-r@v2
         with:
           r-version: ${{ matrix.config.r }}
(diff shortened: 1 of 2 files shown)


.github/workflows/R-CMD-check.yaml
```diff
@@ -20,8 +20,8 @@ jobs:
       matrix:
         config:
           - {os: ubuntu-latest,   r: 'release'}
+          - {os: windows-latest, r: 'release'}
```
roninsightrx, line 23: activated Windows CI. Makes sense here, as it is a user-facing R package.

.github/workflows/R-CMD-check.yaml
```diff
@@ -34,9 +34,6 @@ jobs:
 
       - uses: r-lib/actions/setup-pandoc@v2
       
-      - name: Install curl
```
roninsightrx, line 37: This makes Windows CI fail. It seems like it is an unnecessary library for ubuntu CI, probably a remnant from Jenkins.

PR #102, opened 2024-06-18
add modified wurthwein model of pegasparaginase to PKPDsim models

JordanBrooks33:
Adding pegasparaginase wurthwein to PKPDsim

### DIFF
--- a/inst/models/pk_pegasparaginase_modifiedwurthwein.json5
+++ b/inst/models/pk_pegasparaginase_modifiedwurthwein.json5
@@ -93,7 +93,7 @@
     }
   ],
   "implementations": {
-    "nonmem": "pk_pegasparaginase_wurthwein.mod"
+    "nonmem": "pk_pegasparaginase_modified_wurthwein.mod"
   },
   "comments": [
    "Utilizing updated/simplified model from 2021 publication.",
@@ -105,5 +105,5 @@
     "numerical": true,
     "clinical": false
   },
-  "version": "0.1.3"
+  "version": "0.1.4"
 }
--- a/inst/models/pk_pegasparaginase_modifiedwurthwein.json5
+++ b/inst/models/pk_pegasparaginase_modifiedwurthwein.json5
@@ -1,0 +1,109 @@
+{
+  "id": "pk_pegasparaginase_wurthwein",
+  "description": "PK model for PEG-asparagainase children",
+  "drug": "peg-asparaginase",
+  "category": "oncology",
+  "age_categories": [ "children" ],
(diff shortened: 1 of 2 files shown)


inst/models/pk_pegasparaginase_modifiedwurthwein.json5
```diff
@@ -0,0 +1,109 @@
+{
+  "id": "pk_pegasparaginase_wurthwein",
```
jasmineirx, line 2: We previously discussed using the `modified` tag since this model has been modified from the original. Did you want to include that in this version of the model?

JordanBrooks33: Great catch! Altered.

inst/models/pk_pegasparaginase_modifiedwurthwein.json5
```diff
@@ -93,7 +93,7 @@
     }
   ],
   "implementations": {
-    "nonmem": "pk_pegasparaginase_wurthwein.mod"
+    "nonmem": "pk_pegasparaginase_modified_wurthwein.mod"
   },
```
jasmineirx, line 97: we should remove this field since it isn't useful

inst/models/pk_pegasparaginase_modifiedwurthwein.json5
jasmineirx: we should rename this to be consistent with the file name
```suggestion
  "id": "pk_pegasparaginase_modifiedwurthwein",
```

jasmineirx (changes requested):
Almost there. Can you please refer to one of the other example models in `inst` to ensure only the fields that are present in those models are also in this model? Off the top of my mind, `active` and `build` and `validation` and `admin` are ones that we use for internal tracking, but are not meaningful for the open-source versions of models. I would prefer to keep these files limited to the necessary fields only to aid with readability.

JordanBrooks33:
> Almost there. Can you please refer to one of the other example models in `inst` to ensure only the fields that are present in those models are also in this model? Off the top of my mind, `active` and `build` and `validation` and `admin` are ones that we use for internal tracking, but are not meaningful for the open-source versions of models. I would prefer to keep these files limited to the necessary fields only to aid with readability.

Okay I removed unnecessary fields. Though I didn't see an example with the comments field. I left this as I think it has pertinent information to the modification of the base model cited.

jasmineirx (approved):
lgtm! agree with keeping in the comments since it provides important context

PR #103, opened 2024-06-24
RXR-2140: fix issue due to change in warning behavior in upcoming Rcpp

roninsightrx:
See description of [change in Rcpp](https://github.com/RcppCore/Rcpp/issues/1311#user-content-fn-1-1400efc2b2c5361907ad0f063feeeb20).

The failure in PKPDsim test was due to an erroneously configured test case, which had the wrong number of bins for simulation of inter-occasion variability. This led to extra warnings using the new Rcpp. This PR fixes the test case. 

It also adds an extra warning to make it harder to misconfigure the number of IOV bins (and a new test to confirm the warning is thrown). I did not make that a hard stop(), because in most cases the misconfiguration is not affecting the results.

For reviewing this PR: 
1. first run tests using old version of Rcpp (anything before 1.0.12.4). The tests should all run fine, either on master or on this branch.
2. then install latests version on `Rcpp/main` using `remotes::install("RcppCore/Rcpp")`. Then rerun tests. The tests should fail on the `PKPDsim/master` branch, but not on `PKPDsim/RXR-2140-Rcpp-issue`.

### DIFF
--- a/R/sim.R
+++ b/R/sim.R
@@ -189,6 +189,11 @@ sim <- function (ode = NULL,
         p <- parameters
       }
     }
+    if(!is.null(attr(ode, "iov")$n_bins) && attr(ode, "iov")$n_bins > 1) {
+      if(attr(ode, "iov")$n_bins != (length(iov_bins)-1)) {
+        warning("Number of IOV bins specified for model does not match supplied `iov_bins` argument. This could lead to simulation failures or erroneous output.")
+      }
+    }
     if(is.null(analytical)) {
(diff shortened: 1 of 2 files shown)


jasmineirx (approved):
this looks good to me, but does it make sense for this to be a warning rather than a `stop`? if they are mismatched, something was misspecified by the user

roninsightrx:
> this looks good to me, but does it make sense for this to be a warning rather than a `stop`? if they are mismatched, something was misspecified by the user

I was going back and forth on that, but I think a warning is fine.

PR #105, opened 2024-07-05
RXR-1876 fix warning infusion

roninsightrx:
Refactor of the logic around throwing a warning around infusion rates. This was throwing a harmless but unnecessary warning during automated model validations for some models.

The logic was too strict, it was also thrown for regimens where infusion rates were not relevant. The logic is updated and simplified, as well as the warning text. Also added a test (there was already one test, updated that one too). The warning 
is not thrown any more in the model validation where it occurred before.


jasmineirx (approved):
lgtm!

PR #108, opened 2024-08-26
RXR-2224: fix indexing issue iov bins

roninsightrx:
It was creating an array that was unneccessarily large due to a typo. 
With this update, the behavior is only harmless when the requested number of bins is _lower_ then the number of bins that were originally defined for the model. However, if the number of requested iov bins is _higher_ than defined for the model, it would lead to a runtime error at the C++ level, since it would try to use a vector element that is not defined.
I have changed the behavior now to throw a warning for the harmless case, and an error (at the R level) for the problematic case. Have updated documentation as well. No C++ level errors should occur anymore.

### DIFF
--- a/.github/workflows/pkgdown.yaml
+++ b/.github/workflows/pkgdown.yaml
@@ -20,7 +20,7 @@ jobs:
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
 
       - uses: r-lib/actions/setup-pandoc@v2
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -11,7 +11,7 @@ Authors@R: c(
         person("Jordan", "Brooks", email = "jordan.brooks@insight-rx.com", role = "aut"),
         person("InsightRX", role = c("cph", "fnd")))
 Depends: R (>= 4.0.0)
-Imports: Rcpp (>= 0.12.9), BH, data.table, stringr, MASS,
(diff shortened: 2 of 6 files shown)


roninsightrx:
It seems the failure of CI for pkgdown is due to the bug described [here](https://github.com/pachadotdev/cpp11armadillo/issues/9). I'm not gonna spend time trying to figure out a workaround, will just try to see if it passes next week.

.github/workflows/pkgdown.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
roninsightrx, line 23: unrelated, but they have a newer version of the script available. We're using it in the other action already.

DESCRIPTION
```diff
@@ -11,7 +11,7 @@ Authors@R: c(
         person("Jordan", "Brooks", email = "jordan.brooks@insight-rx.com", role = "aut"),
         person("InsightRX", role = c("cph", "fnd")))
 Depends: R (>= 4.0.0)
-Imports: Rcpp (>= 0.12.9), BH, data.table, stringr, MASS,
+Imports: Rcpp (>= 1.0.13), BH, data.table, stringr, MASS,
```
roninsightrx, line 14: unrelated, saw that we hadn't updated this version requirement (we did in LinkingTo)

mccarthy-m-g (approved):
Looks good!

PR #111, opened 2024-12-06
RXR-2378 bug with Fi defined in ode producing NA concentrations

roninsightrx:
A model where F (bioav) was defined in ODE block as:

```
    F1_avg = F1 * pow(prv_dose/1000.0, -0.15) ;\
    F1i = F1_avg * exp(kappa_F1) ;\
```
resulted in NAs in the simulated concentrations. The code however runs fine when used in PK code.

This was because in the internal C++ code, the first row is not actually a dosing event, but an initalization row that does not have a dose. However, `prv_dose` is initialized as: `prv_dose = doses[0];`, which then results in `prv_dose` being zero. Then in the above definition a value of `prv_dose=0` will results in `Inf` for `F1_avg`.

We can avoid this by instead just reading the first non-zero value in the dose column of the event table and initializing `prv_dose` with that.

This change shouldn't affect anything else.

### DIFF
--- a/inst/cpp/sim.cpp
+++ b/inst/cpp/sim.cpp
@@ -50,18 +50,6 @@ void pk_code (int i, std::vector<double> times, std::vector<double> doses, doubl
   // insert custom pk event code
 }
 
-int find_first_nonzero(std::vector<double> vect) {
-  // find first nonzero value in vector, returns first value if non found
-  int index = 0;
-  for(int i = 0; i < vect.size(); i++) {
-    if(vect[i] > 0) {
-      index = i;
-      break;
-    }
-  }
-  return(index);
-}
-
 // [[Rcpp::export]]
 List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_bins, double step_size) {
   std::vector<double> t;
@@ -93,9 +81,10 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
   // insert A dAdt state_init
   set_covariates(0);
 
-  // set prv_dose to first non-zero dose
-  int idx = find_first_nonzero(doses);
-  prv_dose = doses[idx];
+  // set prv_dose to first dose, if at t=0
+  if(as<NumericVector>(par["nominal_dose_times"])[0] == 0) {
+    prv_dose = as<NumericVector>(par["dose_amts"])[0];
+  }
   t_prv_dose = times[0];
 
   // Main call to ODE solver, initialize any variables in ode code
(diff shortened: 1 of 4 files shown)


inst/cpp/sim.cpp
```diff
@@ -71,7 +83,6 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
   int start;
   memset(rate, 0, sizeof(rate));
   // insert observation compartment
-  // insert bioavailability definition
```
roninsightrx, line 74: Unrelated. This line was actually not used, the grep function was looking for `// insert bioav definition`, so never picked up on this one. A little down in the code it was written correctly. So it seems this line was not functional, therefore can be removed.

inst/cpp/sim.cpp
```diff
@@ -81,7 +92,10 @@ List sim_wrapper_cpp (NumericVector A, List design, List par, NumericVector iov_
   // call ode() once to pre-calculate any initial variables
   // insert A dAdt state_init
   set_covariates(0);
-  prv_dose = doses[0];
+
+  // set prv_dose to first non-zero dose
+  int idx = find_first_nonzero(doses);
```
jasmineirx, line 97: with this change, we are building in a rather quiet assumption that this variable is only correct if doses start at t = 0.

maybe for now this is okay and we can rewrite the model code to something like

```C
F1_avg = F1 \
if (t > 0) { F1_avg = F1 * pow(prv_dose/1000.0, -0.15) } \
```

jasmineirx: On the other hand, other models also could use prv_dose in the ode_code block, and it is maybe also unexpected for it to be 0 momentarily.

jasmineirx (commented):
thanks for looking into this. i didn't realize this was the cause of the issue. a couple thoughts on if it makes sense to code around this or go with the fix.

inst/cpp/sim.cpp
roninsightrx: yes, my thinking was that if we would keep PKPDsim as-is, and have to make some kind of modification in each model, that would be more annoying, and we might forget for future models. I was thinking this modification, but it's similar to yours.
```
FACT = 1
if(prv_dose > 0) FACT = pow(prv_dose/1000.0, -0.15)
F1_avg = F1 * FACT
```
Neither solution is perfect. Let me look into one other solution I was thinking about though, i.e. we could also make sure that `doses[0]` is not zero to begin with (at least as long as at t=0 there is a dose).

roninsightrx:
Update: there is actually another problem that this PR should address. This is that t_prv_dose (the time of previous dose, accessible in ODE block) is currently calculated at the time of the dose + lag time. This is unintuitive, and different from how one would implement it in NONMEM.
TODO: the changes I just put up fixes this behavior somewhat, but is still not entirely correct.

PR #116, opened 2025-05-05
RXR-2577 fix memory access issue 

roninsightrx:
This occurs when specifying IOV bins when not implemented. It will throw a low-level warning, but after that it will also cause a memory access failure that causes R to abort the session.

This PR should fix this, added a test for it as well that failed previously. I've chosen to make this an error and not a warning since this is important to get right, so IMO better than e.g. throwing a soft warning and ignoring the `iov_bins` argument.

### DIFF
--- a/tests/testthat/test_iov.R
+++ b/tests/testthat/test_iov.R
@@ -7,7 +7,21 @@ pars <- list(
   "V" = 50,
   "KA" = 1
 )
-
+pars0 <- list(
+  "CL" = 5,
+  "V" = 50,
+  "KA" = 1
+)
+pk0 <- new_ode_model( # no IOV
+  code = "
+      dAdt[1] = -KA * A[1]
+      dAdt[2] = +KA * A[1] -(CL/V) * A[2]
+    ",
+  obs = list(cmt = 2, scale = "V"),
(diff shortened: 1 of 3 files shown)


tests/testthat/test_iov.R
```diff
@@ -7,7 +7,21 @@ pars <- list(
   "V" = 50,
   "KA" = 1
 )
-
+pars0 <- list(
+  "CL" = 5,
+  "V" = 50,
+  "KA" = 1
+)
+pk0 <- new_ode_model( # no IOV
```
jasmineirx, line 15: calling new_ode_model is slow, and I don't think we strictly need this model for this test. we load a number of pretty basic models in our testing suite, can we use one of those to keep test time down?

jasmineirx (commented):
fix looks good to me, but one question about testing

tests/testthat/test_iov.R
roninsightrx: Yes, I tried, but couldn't find a good one while having the test be is in a logical test file.
 
What I could do is make a separate ticket to restructure the tests to use only the minimal number of calls to new_ode_model(), I think there is some unnecessary duplication going on as well.

PR #117, opened 2025-05-14
RXR-2590: add support for additional "types" of oral dosing

jasmineirx:
`new_regimen()` will now use regex to look for the word "oral" instead of requiring an exact match to the word "oral", allowing the same t_inf adjustments for any type with "oral" in it (e.g., "oral_tablet", "oral_suspension").

I also noticed that we had conflicting logic for SC and IM dosing, where in `new_regimen()` we explicitly test for non-0 infusions for SC while in `create_event_table()` we forced these to be 0. I resolved this conflict by allowing SC and IM dosing to have non-zero t_inf.

### DIFF
--- a/R/create_event_table.R
+++ b/R/create_event_table.R
@@ -120,9 +120,9 @@ create_event_table <- function(
   }
 
   # parse list to a design (data.frame)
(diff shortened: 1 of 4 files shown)


roninsightrx (approved):
lgtm!

PR #118, opened 2025-05-22
RXR-2593: Ensure covariate times are correct when t_init is non-zero

jasmineirx:
We allow the argument t_init to be specified in the case of observations prior to the first dose. When this occurs, we shift the regimen start time to begin at t_init, effectively changing the "origin". However, we were not also shifting covariate times. As a result, there could be a mismatch in covariate times and dose times in simulation.

I opted to call PKPDsim::new_covariate again to update the covariate times since this ensures the covariate values are correctly extrapolated backwards to t = 0, respecting existing implementation method for extrapolation.


roninsightrx (approved):
lgtm, this indeed seems like an oversight not to adapt the covariates.
Make sure also the tests in PKPDmap still pass CI, there are some that use t_init.

PR #121, opened 2025-06-24
fix issue with merge based on floating points

jasmineirx:
It was possible to get an issue when we merge together observation data and covariate data due to floating point comparison issues.

We now merge on a time column that has been rounded to 6 decimal places, which is consistent with behavior elsewhere in the package.


mccarthy-m-g (approved):

roninsightrx (approved):
ok

PR #122, opened 2025-07-07
RXR-2394: lagtime updates

roninsightrx:
Lagtime was previously handled by just adding the lagtime for a given compartment to each dose given into that compartment. So on the level of the event table created that was used as input for the simulation. However, this introduced the problem that the lagtime could not be estimated by PKPDmap (because it is essentially fixed). 
To solve this, I moved the functionality that adapted the event table from happening before the core simulation into the core simulation function itself, right before the actual simulation. Now, this core function requires an additional lagtime parameter, which is then used to adapt the event table.

Related PR in [PKPDmap](https://github.com/InsightRX/PKPDmap/pull/51)


R/compile_sim_cpp.R
```diff
@@ -270,12 +270,52 @@ compile_sim_cpp <- function(
   if(length(grep("-w", flg)) == 0) {
     Sys.setenv("PKG_CXXFLAGS" = paste(flg, "-w"))
   }
+
   if(compile) {
-    Rcpp::sourceCpp(code=sim_func, rebuild = TRUE, env = globalenv(), verbose = verbose, showOutput = verbose)
+    Rcpp::sourceCpp(
```
roninsightrx, line 275: just aesthetic change

inst/cpp/sim.cpp
```diff
@@ -63,23 +63,116 @@ void pk_code (int i, std::vector<double> times, std::vector<double> doses, doubl
   // insert custom pk event code
 }
```
roninsightrx, line 65: this file is becoming fairly large. In principle we should separate off into modular design. However, some issues with that, see [RXR-2712](https://insight-rx.atlassian.net/browse/RXR-2712). We should tackle in separate PR.

inst/cpp/sim.cpp
```diff
@@ -63,23 +63,116 @@ void pk_code (int i, std::vector<double> times, std::vector<double> doses, doubl
...
-  evid = as<std::vector<int> >(design["evid"]);
-  dummy = as<std::vector<double> >(design["dum"]);
-  rates = as<std::vector<double> >(design["rate"]);
-  dose_cmt = as<std::vector<int> >(design["dose_cmt"]);
-  dose_type = as<std::vector<int> >(design["type"]);
-  obs_type = as<std::vector<int> >(design["obs_type"]);
+
+  // Handle lagtime parameter - can be numeric or character
+  NumericVector lagtime_numeric = lagtime_to_numeric(lagtime, par);
+  List events = apply_lagtime(design, lagtime_numeric, n_comp);
```
roninsightrx, line 166: first convert the vector to numeric (e.g. from possible parameter "TLAG"), then apply the lagtime vector to the event table.

inst/cpp/sim.cpp
```diff
@@ -63,23 +63,116 @@ void pk_code (int i, std::vector<double> times, std::vector<double> doses, doubl
...
+  
+  times = as<std::vector<double> >(events["t"]);
```
roninsightrx, line 168: just working off of the new event table, otherwise unchanged.

roninsightrx (commented):
few comments

R/parse_lagtime.R
```diff
@@ -0,0 +1,20 @@
...
+) {
+  if(!is.null(attr(ode, "lagtime")) && attr(ode, "lagtime")[1] != "undefined" && attr(ode, "lagtime")[1] != "NULL") {
+    if(is.null(lagtime)) { # only override from metadata if not specified by user
+      lagtime <- attr(ode, "lagtime")
+    }
+  }
+  if(is.null(lagtime)) {
+    lagtime <- c(0)
+  }
+  lagtime
```
jasmineirx, line 19:
```suggestion
  lagtime_ode <- attr(ode, "lagtime")
  # override from ode if not specified by user and defined in ode
  if(is.null(lagtime) && !is.null(lagtime_ode) && !lagtime_ode %in% c("NULL", "undefined")) {
    lagtime <- lagtime_ode
  }
  if(is.null(lagtime)) {
    lagtime <- 0
  }
  lagtime
```

might be slightly more readable?

R/sim.R
```diff
@@ -602,6 +598,12 @@ sim <- function (ode = NULL,
   all_names <- unique(c(par_names, cov_names, var_names))
   all_names <- intersect(all_names, names(comb)) # only cols that appear in data
 
+  ## remove dose-times from regimen
+  ## but leave ones that were also in t_obs
+  ## also leave extra obs when bolus, because this may be needed (controllec below using `extra_t_obs`)
```
jasmineirx, line 603:
```suggestion
  ## also leave extra obs when bolus, because this may be needed (controlled below using `extra_t_obs`)
```
typo

inst/cpp/sim.cpp
```diff
@@ -63,23 +63,116 @@ void pk_code (int i, std::vector<double> times, std::vector<double> doses, doubl
...
+  } else if (TYPEOF(lagtime) == STRSXP) {
+    CharacterVector lagtime_char = as<CharacterVector>(lagtime);
+    lagtime_numeric = NumericVector(lagtime_char.size());
+    for (int i = 0; i < lagtime_char.size(); i++) {
+      String param_name = lagtime_char[i];
+      if (parameters.containsElementNamed(param_name.get_cstring())) {
+        lagtime_numeric[i] = as<double>(parameters[param_name]);
+      } else {
+        lagtime_numeric[i] = 0.0; // default value
+      }
```
jasmineirx, line 79: am i correct that this means that if we supply lagtime as a variable (e.g. "ALAG") but the supplied parameters do not contain that variable, we quietly override with zero? should we implement a check for this on the R side maybe?

inst/cpp/sim.cpp
```diff
@@ -63,23 +63,116 @@ void pk_code (int i, std::vector<double> times, std::vector<double> doses, doubl
...
+      lagtime[i] = 0.0;
+    }
+  } else {
+    lagtime = tlag;
+  }
+
+  // Apply lagtime to dose events (evid == 1)  
+  for(int i = 0; i < times.size(); i++) {
+    if(evid[i] == 1) {
+      times[i] += lagtime[cmt[i]-1];
```
jasmineirx, line 111: just confirming the `-1` is because cmt is 1-indexed and not 0-indexed?

inst/cpp/sim.cpp
jasmineirx: where do we pull n_comp from?

tests/testthat/test_sim_core.R
```diff
@@ -16,3 +16,47 @@ test_that("sim core works", {
...
+    obj <- sim(
+      ode = mod_1cmt_oral_lagtime, 
+      regimen = reg, 
+      parameters = par, 
+      only_obs = TRUE, 
+      t_obs=c(0:24), 
+      return_design=TRUE
+    )
+    sim_core(obj, ode = mod_1cmt_oral_lagtime)$y
+  }
```
jasmineirx, line 45: we aren't using this one, is that an oversight?

jasmineirx (commented):
in general looks good! a couple minor comments

inst/cpp/sim.cpp
roninsightrx: it's a global variable that's added dynamically before the file gets compiled. Not ideal that it is global, but we have a ticket to address ([RXR-2712](https://insight-rx.atlassian.net/browse/RXR-2712).)

inst/cpp/sim.cpp
roninsightrx: yes, that's how it works. Good point, I'll add a warning on R side

jasmineirx (approved):

PR #123, opened 2025-08-12
RXR-2789: also adjust covariate times to account for prepended steady-state dosing

jasmineirx:
This PR ensures we correctly adjust t_obs and covariate times for both t_init and t_ss. Previously, we were only adjusting for t_init.

I made more lines of code changes than strictly necessary so that it is easier to track which time-containing variables are adjusted when. We had been inconsistently adding t_init and t_ss, and now we should generally be adjusting for one or both at the same time, as appropriate.

I opted to keep t_init and t_ss as separate variables rather than lumping them together as one variable (e.g., t_rezero or something) because it allows for better partitioning of the sources of time adjustments. we do handle them separately in a couple places.

I added some comments throughout to clarify locations that I found confusing. I also added quite a few unit tests for various combos of t_ss and t_init and covariates/covariates_table

### DIFF
--- a/R/create_event_table.R
+++ b/R/create_event_table.R
@@ -196,7 +196,10 @@ create_event_table <- function(
       t_tte <- t_tte + t_init
       if(is.null(t_tte) || is.na(t_max) || t_max < max(t_tte)) { t_max <- max(t_tte) }
     }
+  } else if (!is.null(t_obs)) {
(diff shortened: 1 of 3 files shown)


mccarthy-m-g (approved):
Nice work! Well-tested and the code comments are a helpful addition

roninsightrx (approved):
lgtm, well-tested.

PR #124, opened 2025-08-20
RXR-2811: fix time shift bug introduced in RXR-2789

mccarthy-m-g:
This PR fixes a small bug introduced in https://github.com/InsightRX/PKPDsim/pull/123 for steady-state regimens where t_init != 0 and there was an observation before t == 0. We were filtering these observations out of the output, but we want to retain them.

### DIFF
--- a/R/sim.R
+++ b/R/sim.R
@@ -138,7 +138,6 @@ sim <- function (ode = NULL,
   # to shift dosing only as much as we need to. Keep t_init and t_ss as separate
   # variables, however, since we need to refer to these time shifts in different
   # places.
-  t_init_orig <- t_init
   t_init <- pmax(0, t_init - t_ss)
   # adjust max simulation time based on time shifts + specified obs
   if (!is.null(t_max)) t_max <- t_max + t_init + t_ss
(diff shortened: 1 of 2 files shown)


R/sim.R
```diff
@@ -631,7 +632,7 @@ sim <- function (ode = NULL,
   if(!is.null(regimen_orig$ss_regimen)) {
     t_ss <- utils::tail(regimen_orig$ss_regimen$dose_times,1) + regimen_orig$ss_regimen$interval
     comb$t <- as.num(comb$t) - t_ss
-    comb <- comb[comb$t >= 0,]
+    comb <- comb[comb$t >= pmin(0, t_init_orig - t_ss),]
```
jasmineirx, line 635: I think this could use a code comment for why t_init_orig instead of t_ss

jasmineirx (approved):
lgtm, one small suggestion. nice fix.

R/sim.R
mccarthy-m-g: We determined that this line is not necessary because times are still shifted by t_ss and/or t_init here when we extract them, therefore we don't need to worry about steady state doses being included in the output:

https://github.com/InsightRX/PKPDsim/blob/b32a41b9e91b41e74d768afda10d31ff422d808f/R/sim.R#L628

jasmineirx (approved):

PR #125, opened 2025-08-20
RXR-2719: cleanup: update docs; rm ellipsis

roninsightrx:
- improve description of argument
- remove ellipsis, it wasn't used in the code

### DIFF
--- a/R/sim.R
+++ b/R/sim.R
@@ -11,7 +11,11 @@
 #'   population PK parameters are supplied, a value of `1` or `2` should be
 #'   passed, since currently only two mixture groups are supported.
 #' @param omega vector describing the lower-diagonal of the between-subject variability matrix
-#' @param omega_type exponential or normal, specified as vector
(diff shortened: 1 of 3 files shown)


tests/testthat/test_get_var_y.R
```diff
@@ -111,7 +111,7 @@ test_that("Two compartment model", {
     parameters = par2,
     t_obs = t_obs,
     regimen = reg,
-    mega = omega2
+    omega = omega2
```
roninsightrx, line 114: actually because of the ellipsis this wasn't caught before. The code ran, but the omega argument wasn't specified so the computations were wrong. We just had ridiculously high tolerance on the test values, so we didn't see it.

mccarthy-m-g (approved):
nice cleanup! 🧹

roninsightrx:
PKPDmap unit tests are failing, need some updates there too due to ellipsis being removed.

PR #126, opened 2025-09-05
fix thomson and install function

roninsightrx:
- Thomson has 3 fixed parameters, should be only 1. This throws an error in get_map_estimates() currently. Since we remove TDM_INIT, we also need to switch this model to not use an init parameter. Also fixed this same issue for two other models.

- The function install_default_litreature_model() is not passing … along to model_from_api(). It says in documentation it should (and I also think it should)

None of this should affect any production code, this is all example code / models.

### DIFF
--- a/inst/models/pk_vanco_colin.json5
+++ b/inst/models/pk_vanco_colin.json5
@@ -50,8 +50,7 @@
     0.684
   ],
   "fixed": [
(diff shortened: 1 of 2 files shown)


jasmineirx (approved):
lgtm

PR #128, opened 2025-11-14
Add compartments cmt to regimens in nm_to_regimen

dominic-irx:
This PR changes nm_to_regimen so that it uses a map for compartment to dose type. This is useful for mipdeval, where this is needed.

This breaks backward compatibility because we now want an explicit declaration of the type of dosing instead of inferring from presence/absence of columns.

This function is used in two places:

1. PKPDposterior::nonmem_to_stan_data:40, which we haven't touched in a while
2. mipdeval::parse_input_data:45, which is where we have the temporary fix

### DIFF
--- a/tests/testthat/test_nm_to_regimen.R
+++ b/tests/testthat/test_nm_to_regimen.R
@@ -103,3 +103,21 @@ test_that("Doses in different compartments handled", {
   expect_equal(reg2$t_inf, c(1, 1, 1, 0.5, 0, 0))
   expect_equal(reg2$cmt, c(2, 2, 2, 2, 1, 1))
 })
+
+test_that("Different rates handled correctly", {
+  pt2 <- data.frame(
+    ID = 1,
+    EVID = c(1, 1, 0),
+    CMT = 2, 
+    AMT = c(100, 200, 0),
+    TIME = c(0, 6, 23),
+    RATE = c(0, 100, 0),
+    DV = c(0, 0, 5)
+  )
+  reg2 <- nm_to_regimen(pt2)
+  expect_true(inherits(reg2, "regimen"))
+  expect_equal(reg2$dose_amts, c(100, 200))
+  expect_equal(reg2$dose_times, c(0, 6))
+  expect_equal(reg2$t_inf, c(0, 2))
+  expect_equal(reg2$cmt, c(2, 2))
+})


R/nm_to_regimen.R
```diff
@@ -19,29 +23,44 @@ nm_to_regimen <- function(data,
...
       tmp$time <- tmp$time - min(tmp$time)
     }
-    if(type == "infusion") {
-      reg[[i]] <- new_regimen(amt = tmp$amt, times = tmp$time, type = "infusion", t_inf = tmp$amt / tmp$rate)
+    if (!is.null(dose_cmts)){
+      # map cmt to dose type
+      reg[[i]] <- new_regimen(
+        amt = tmp$amt,
+        times = tmp$time,
+        cmt = tmp$cmt,
```
jasmineirx: I think we could skip this argument. if it is null, `PKPDsim::sim` will use the compartment definitions of the model file, which is probably safer.

jasmineirx: Could you help me understand this case with an example? It is possible that NONMEM-style data and PKPDsim model definitions use different compartments for one dose type.

dominic-irx: https://github.com/InsightRX/mipdeval/pull/25/files

This is the **only** relevant place this function is currently being used, where we want to specify which compartments are from which doses using the NONMEM data set. Leaving it to be assumed by the model downstream, especially when CMT should be specified in the data set anyway, is not optimal imo.

R/nm_to_regimen.R
```diff
@@ -19,29 +21,57 @@ nm_to_regimen <- function(data,
...
+        # if no RATE is given, then assume bolus
+        suppressWarnings({
+          reg[[i]] <- new_regimen(
+            amt = tmp$amt,
+            times = tmp$time,
+            cmt = tmp$cmt
+          )
+        })
+      }
+    } else if ("rate" %in% colnames(doses) &! 0 %in% doses$rate){
```
roninsightrx: What if we have two doses: on a first dose RATE=0 (bolus loading) and on second dose RATE=1? Then all doses will be treated as bolus, I think. Would add that as a test case to check, and potentially update behavior?

roninsightrx (commented):
looks good, but one last question about edge case

roninsightrx (approved):
lgtm

PR #129, opened 2025-11-21
RXR-2907: add new models for open access, fix misnamed fields

jasmineirx:
We recently wrote a manuscript reporting model performance across a few adult models. We should make this code available for transparency. This PR adds the model specification used for these models.


JordanBrooks33 (approved):
lgtm

PR #135, opened 2026-03-23
RXR-3049: small bug when interval = infusion length

jasmineirx:
Fixes a bug where we tried to assign values to a row of length 0, in some cases (interval = infusion length); also removes mockery from tests (deprecated package)

### DIFF
--- a/tests/testthat/test_calc_auc_analytic.R
+++ b/tests/testthat/test_calc_auc_analytic.R
@@ -113,6 +113,25 @@ test_that("Works for 1cmt model", {
   expect_equal(tmp$auc, aucfr$auc)
 })
 
+test_that("Doesn't crash when t_inf equals interval (all infusion-end times coincide with existing data times)", {
+  # Regression test: when t_inf = interval, every infusion-end event lands on the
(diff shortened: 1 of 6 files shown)


tests/testthat/test_calc_auc_analytic.R
```diff
@@ -113,6 +113,25 @@ test_that("Works for 1cmt model", {
   expect_equal(tmp$auc, aucfr$auc)
 })
 
+test_that("Doesn't crash when t_inf equals interval (all infusion-end times coincide with existing data times)", {
```
mccarthy-m-g:
```suggestion
test_that("No error when t_inf equals interval (all infusion-end times coincide with existing data times)", {
```

mccarthy-m-g (approved):
lgtm! Just a couple tidy-up suggestions:

mccarthy-m-g (approved):
New changes look good! Hopefully you don't find a third bug... Just a small grammar suggestion for a code comment that took me a bit to understand. 

Do you need to check f-tests before merging?

PR #137, opened 2026-05-08
RXR-3133: add flexibility in how we determine infusion

jasmineirx:
In several places in the codebase, we need to determine if something is an "infusion-type" dose or an "oral/bolus type" dose. See, for example,

https://github.com/InsightRX/PKPDsim/blob/b71e52186d6d8bc04d63785bbe70514fde82a19a/R/create_event_table.R#L125

In the advan/analytic equations, we were hard-coding looking for the exact string "infusion", which breaks when other "infusion type" doses are used. I modified the data creation function to use the same logic we use elsewhere (i.e., look for the string "oral" in the "type" or look for the exact word "bolus"


mccarthy-m-g:
It might be worth turning the `is_bolus` logic into a function, now that we're using it in two places.

```r
is_bolus_or_oral <- function(regimen) {
   regimen$type == "bolus" | grepl("oral", regimen$type)
}
```

And then update the line here:

https://github.com/InsightRX/PKPDsim/blob/b71e52186d6d8bc04d63785bbe70514fde82a19a/R/create_event_table.R#L125

to something like:

```r
is_bolus <- regimen$type == "covariate" | is_bolus_or_oral(regimen)
```

jasmineirx:
> ```r
> is_bolus_or_oral <- function(regimen) {
>    regimen$type == "bolus" | grepl("oral", regimen$type)
> }
> ```

Good suggestion. I made the helper function and applied to to all the other related places I could find in the code base. Note that this does cause one change in functionality in that anything that is false for is_bolus_or_oral will now have "infusion logic" applied for minimum infusion time, which I believe is desired behaviour. For example, before we would have treated infusion_prodrug as being allowed to have t_inf of zero, when I think we would want this to be handled as any other infusion-type dose.

mccarthy-m-g (approved):
> I made the helper function and applied to all the other related places I could find in the code base. Note that this does cause one change in functionality in that anything that is false for is_bolus_or_oral will now have "infusion logic" applied for minimum infusion time, which I believe is desired behaviour. For example, before we would have treated infusion_prodrug as being allowed to have t_inf of zero, when I think we would want this to be handled as any other infusion-type dose.

Changes look good! Are there any bolus type doses that will be affected by this? I'm pretty sure we only use/need `"bolus"`, but just double-checking. If we think we'll ever use other labels for bolus type doses we could use grepl in `is_bolus_or_oral()` for bolus too, instead of strict equality

Do any f-tests need updating with this change?

jasmineirx:
> > I made the helper function and applied to all the other related places I could find in the code base. Note that this does cause one change in functionality in that anything that is false for is_bolus_or_oral will now have "infusion logic" applied for minimum infusion time, which I believe is desired behaviour. For example, before we would have treated infusion_prodrug as being allowed to have t_inf of zero, when I think we would want this to be handled as any other infusion-type dose.
> 
> Changes look good! Are there any bolus type doses that will be affected by this? I'm pretty sure we only use/need `"bolus"`, but just double-checking. If we think we'll ever use other labels for bolus type doses we could use grepl in `is_bolus_or_oral()` for bolus too, instead of strict equality
> 
> Do any f-tests need updating with this change?

Our end-to-end tests run fine with this change. There are no derivative bolus types that we use.

PR #138, opened 2026-06-09
RXR-3174: safety check to ensure max sim time is reasonable

jasmineirx:
When trying different parameter estimates during MAP Bayesian optimization, it is possible to achieve a TLAG estimate that is very very large. In this case, it is possible that the vector of timepoints simulated becomes very, very large.

In theory, we shouldn't ever need to simulate past the t_obs time point.

This PR addresses the TLAG large vector issue by limiting sim time. Tests in this repo and dependencies pass.


roninsightrx (approved):
lgtm

mccarthy-m-g (approved):

PR #141, opened 2026-06-25
Update logic for setting rate in model, when exporting regimen to NONMEM-style data.frame

roninsightrx:
This fixes the behavior in the function that created NONMEM-style datasets from PKPDsim regimens. The earlier [modified behavior](https://github.com/InsightRX/PKPDsim/commit/399acbfa3cb3de158de94c471597bf6ab28b6622) was implemented to support models with infusion rates AND bioavailability specified. I believe the use-case for this was some oral models with zero-order absorption. In those cases, we need to make a small adjustment to the NONMEM model, and specify rate in the model itself, to properly account for bioavailability. However, for the majority of models (iv infusions) this is not needed. If `bioav` was set to 1 the logic shouldn't fire, and rate shouldn't have to be defined in the NONMEM model. This logic was now added.

**Note**: This PR doesn't touch any production code, it is just in a convenience function that is used in irxmodval.

### DIFF
--- a/R/regimen_to_nm.R
+++ b/R/regimen_to_nm.R
@@ -1,8 +1,8 @@
 #' Convert PKPDsim regimen to NONMEM table (doses only)
 #'
-#' Note: when bioavailability is used for insusions: NONMEM behaves differently from PKPDsim and Monolix, 
-#' in that rates are not automatically recomputed for infusions where bioavailabiliy also applies. 
-#' In PKPDsim/Monolix, for a bioavailability of 50%, an AMT of 100 mg and rate of 100mg/hr would be recalculated 
+#' Note: when bioavailability is used for insusions: NONMEM behaves differently from PKPDsim and Monolix,
+#' in that rates are not automatically recomputed for infusions where bioavailabiliy also applies.
+#' In PKPDsim/Monolix, for a bioavailability of 50%, an AMT of 100 mg and rate of 100mg/hr would be recalculated
 #' to 50 mg and 50 mg/hour, respectively, to keep the infusion length the same. This is not the case
 #' in NONMEM. Therefore, the easiest way to work around this is for the user to compute the rate manually
 #' in NONMEM using custom code. Therefore, we set the `RATE` column to `-1` for such infusions.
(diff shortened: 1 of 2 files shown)


roninsightrx:
Confirmed this fiixes issue with model validation:

before fix (master):
- pk_atg_ghazal: passes
- pk_vanco_mccarthy: fails

after fix (this branch):
- pk_atg_ghazal: passes
- pk_vanco_mccarthy: passes

R/regimen_to_nm.R
```diff
@@ -1,8 +1,8 @@
 #' Convert PKPDsim regimen to NONMEM table (doses only)
 #'
-#' Note: when bioavailability is used for insusions: NONMEM behaves differently from PKPDsim and Monolix, 
-#' in that rates are not automatically recomputed for infusions where bioavailabiliy also applies. 
-#' In PKPDsim/Monolix, for a bioavailability of 50%, an AMT of 100 mg and rate of 100mg/hr would be recalculated 
+#' Note: when bioavailability is used for insusions: NONMEM behaves differently from PKPDsim and Monolix,
```
mccarthy-m-g:
```suggestion
#' Note: when bioavailability is used for infusions: NONMEM behaves differently from PKPDsim and Monolix,
```

R/regimen_to_nm.R
```diff
@@ -50,8 +50,13 @@ regimen_to_nm <- function(
         }
         bioav_dose[is.na(bioav_dose)] <- 1
       }
-      dat$RATE <- -1
-      message("Setting rate to be handled in NONMEM model using R parameters.")
+      ## Only set rate to model-specified if bioav != 1
+      ## For many IV models, bioav may be set explicitly to 1.
+      ## But there we just want to ignore.
+      if(! isTRUE(as.numeric(bioav[dose_cmt]) == 1)) {
```
mccarthy-m-g:
```suggestion
      if(!isTRUE(as.numeric(bioav[dose_cmt]) == 1)) {
```

mccarthy-m-g (approved):
lgtm! I also checked pk_posaconazole_chen, which has a mix of oral (bioav: 0) and iv (bioav: 1) and that was fixed too, so I think our bases are covered for where this was breaking. Just a couple grammar/formatting suggestions below:

PR #145, opened 2026-08-31
Mark '^' fixed in error branch

### LINKED ISSUE #144: Fixed string assumed in regex will always pass (or error)
https://github.com/InsightRX/PKPDsim/blob/99f8ea552078f086187083bf581f759aef783bed/R/translate_ode.R#L71

Here, `^` is a regular expression -- it matches every string but `NA` (https://github.com/gagolews/stringi/issues/529). It would appear `fixed("^")` is intended.

MichaelChirico:
Closes #144


jasmineirx (approved):
lgtm, thanks for the contribution.

